PR #185, opened 2020-04-16
Layout upgrades

nulinspiratie:
Used to be PR #181 before the repository cleaning

List of upgrades:
- Add function `Layout.plot_traces`, which plots the traces (before segmenting per pulse)
- Change `layout.sample_rate` from an `@property` into a Parameter
- `layout.is_acquiring` parameter, to ensure that two acquisitions aren't performed simultaneously
- Add `layout.timings`, which is a `PerformanceTimer` that records how long it takes to setup, acquire etc.
- Do not copy pulse when adding to each interface during targeting. Adds a slight speedup
- Add `pulse_slices` and `pulse_slices_full` as metadata to traces


silq/meta_instruments/layout.py
```diff
@@ -1587,28 +1656,80 @@ def save_traces(self,
...
+        for k, (ax, (channel, traces_arr)) in enumerate(zip(plot, traces.items())):
+                if traces_arr.shape[0] == 1:
+                    ax.add(traces_arr[0], x=t_list, **plot_kwargs)
+                    ax.set_ylabel('Amplitude (V)')
+                else:
+                    ax.add(traces_arr, x=t_list, y=sample_list, **plot_kwargs)
+                ax.set_ylabel('Sample number')
+                ax.set_xlabel('Time (ms)')
+                ax.set_title(channel)
+
```
maij, line 1729: Could we make `DataArray`s here so the units are handled automatically in MatPlot?

maij (commented):
Just 1 comment regarding plot_traces, I'm happy for this to be merged soon.

silq/meta_instruments/layout.py
nulinspiratie: Good idea! I added it and looks good

maij (approved):

PR #191, opened 2020-04-16
feature/add t_read to blips_parameter

nulinspiratie:


nulinspiratie:
Might be worthwhile adding later, but currently a distraction

PR #194, opened 2020-04-16
circuit tools

nulinspiratie:
TODO
- [x] Add regex package requirement
- [x] Remove `NMRCircuitPulseSequence`, as `CircuitPulseSequence` has the same usage and is more general
- [x] Either remove `NMRCircuitParameter`, or modify it to use `CircuitPulseSequence`

Originally #158 before repo clean


silq/analysis/analysis.py
```diff
@@ -117,8 +201,9 @@ def find_high_low(traces: np.ndarray,
 
     # Find two peaks by changing the threshold dependent on the number of peaks foudn
```
maij, line 202:
```suggestion
    # Find two peaks by changing the threshold dependent on the number of peaks found
```

maij (changes requested):
This PR should be split into multiple PRs, I think I've seen some of these changes elsewhere. 
@RostyslavSavytskyy, looking at files like the layout, analysis, and acquisition parameters do you know where these changes originated from?

The majority of this PR is not at all about circuit tools.

nulinspiratie:
Strange, it looks as though it is implementing changes that were already merged into silq in previous PR's. Perhaps master needs to be merged into this branch first?

silq/tools/circuit_tools.py
```diff
@@ -0,0 +1,162 @@
...
+                expanded_circuit.append(gate.name)
+            elif str(gate) == '[]':
+                gate.append('Gi')
+            else:
+                raise RuntimeError(f'Cannot understand gate {gate}')
+
+    elif isinstance(circuit, str):
+        # Replace [] by identity gate
+        expanded_circuit = circuit.replace('[]', 'Gi')
+
```
RostyslavSavytskyy, line 25: would be nice to add here option for Randomized Benchmarking circuits, where we need to remove qubit name, i.e. something like: 
expanded_circuit = circuit.replace(f'{q_name}', '')

thoughts, opinions?

RostyslavSavytskyy:
I pull the master branch, so now it is as it should be

silq/tools/circuit_tools.py
nulinspiratie: Can you please give a concrete example? I don't recall anymore how the circuit refers to specific qubit names.
just on first thought I think as soon as there's more than 1 qubit it would be useful if the gates specify the relevant qubit

nulinspiratie:
Hey @RostyslavSavytskyy thanks for that. the branch looks much cleaner now.

There's (at least) one more modification which needs to happen in this PR before I think it's ready to be merged. I've added this as a tickbox in the description, but basically we should delete `NMRCircuitPulseSequence` and modify `NMRCircuitParameter` to use `CircuitPulseSequence` instead. I'll see when I have time for this, but else if someone is keen feel free to take that on. It needs to be tested on a setup anyway.

I've also made one minor commit such that we no longer need pygsti as a package requirement (which would be a pain).

silq/tools/circuit_tools.py
RostyslavSavytskyy: Yes, so if you generate circuits in pygsti for Randomized Benchmarking, you need to define the name of the qubit and hence using 'Q' as an example will give you a circuit like this: "Gxmpi2:QGympi2:QGypi2:QGxpi2:Q@(Q)  0  0"
Obviously if we do for two or more qubits situation goes to Q1, Q2, etc.

nulinspiratie: I think in the general case it is important to keep the qubit name, since you need a way to specify what qubit a pulse should be applied to when there is more than one qubit. 

When there is only one qubit, you could argue that the `":Q"` is unnecessary. We could add a renaming to remove this. But if we implement this, there are two complications:
1. It should also check if the pulse with `":Q"` exists, because the user might decide to add that at the end of the pulse name.
2. If we remove `":Q"`, we need to be careful that we only do this if no number follows the `Q`, because otherwise it ignores what qubit the pulse should be applied to.

One more thing to note is that if you define your pulse in `settings.pulses` as e.g. `"Gxmpi2:Q"` everything should be fine. So this implementing this is not a necessity, though it could be nice. At the moment I don't have time to add this. If anyone else is keen please feel free to do so. Otherwise I suggest we change this to an Issue, and add it in a later PR.

P.S. defining gates in a circuit with `:Q` is not just specific to RB, but also to the newer GST implementations

RostyslavSavytskyy: I like the idea of defining gates/pulses as "Gxmpi2:Qn", where n can be anything. This looks quite general and there is no need to complicate things, just need to adhere everyone to one convention

RostyslavSavytskyy:
Not sure whether I did it correctly (needs testing), but made the above changes with CircuitPulseSequence

nulinspiratie:
Alright I think it's in a good enough stage to be merged. I'll try to find the time to add a few docstrings.
What do you guys think @maij @RostyslavSavytskyy ?

RostyslavSavytskyy (approved):
Looks good to me, can approve

silq/tools/circuit_tools.py
```diff
@@ -0,0 +1,175 @@
...
+        load_probabilities: bool = False
+):
+    filepath = Path(filepath)
+    with open(filepath.with_suffix('.txt')) as file:
+        lines = [line.strip() for line in file]
+
+    # Check if line already contains data
+    if lines[0].startswith('##'):
+        # Remove two headers
+        lines = lines[1:]
```
maij, line 101: Looks like you're removing a single header here.

silq/tools/circuit_tools.py
```diff
@@ -0,0 +1,175 @@
...
+    # Ensure all strings are of the target type
+    circuits = [
+        convert_circuit(circuit, target_type=target_type)
+        for circuit in circuit_strings
+    ]
+
+    if load_probabilities:
+        # Ignore when there is more than one space
+        lines = [line.replace('   ', ' ') for line in lines]
+        lines = [line.replace('  ', ' ') for line in lines]
```
maij, line 115: Alternatively, using regex
```suggestion
        lines = [re.sub(' +', ' ', line) for line in lines]
```

maij (changes requested):
Small changes requested.

silq/tools/circuit_tools.py
maij: Hmm, this is ever so slightly slower than the string replace method.
Arguably, it reads better than the 2 lines. Up to you guys in the end.

Final alternative which I just remembered works: `lines = [' '.join(line.split()) for line in lines]`

silq/tools/circuit_tools.py
nulinspiratie: fixed

silq/parameters/acquisition_parameters_composite.py
```diff
@@ -390,3 +391,117 @@ def plot_flips(self , figsize=(8, 3)):
...
+class NMRCircuitParameter(NMRParameterComposite):
+    """ Parameter for most measurements involving an NMR pulse.
+
+    This parameter can apply several NMR pulses, and also measure several ESR
+    frequencies. It uses the `NMRPulseSequence`, which will generate a pulse
+    sequence from settings (see parameters below).
+
+    In general, the pulse sequence is as follows:
+
+    1. Perform any pre_pulses defined in ``NMRParameter.pre_pulses``.
```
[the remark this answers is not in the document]
nulinspiratie: Fixed. I've now based it on the docstring of `NMRParameterComposite`

silq/tools/circuit_tools.py
nulinspiratie: Nice, I like it, fixed

maij (approved):

PR #200, opened 2020-04-17
Feature/improved loading of pulse sequence

nulinspiratie:
Add an additional check when loading a pulse sequence from a pickle


maij:
Any idea why some pickled files are unreadable?

nulinspiratie:
Nope 🥇

PR #201, opened 2020-04-17
added nuclear state probability/proportion analysis to NMR parameter

RostyslavSavytskyy:
straightforward calculation of nuclear state probability in NMR analysis

### DIFF
--- a/silq/parameters/acquisition_parameters.py
+++ b/silq/parameters/acquisition_parameters.py
@@ -1936,7 +1936,7 @@ class NMRParameter(AcquisitionParameter):
 
             state_probability_0[f_idx] = np.sum(up_proportions[f_idx] <
                                                 self.threshold_up_proportion)/self.samples
-            state_probability_1[f_idx] = np.sum(up_proportions[f_idx] >
+            state_probability_1[f_idx] = np.sum(up_proportions[f_idx] >=
                                                 self.threshold_up_proportion)/self.samples
 
             if len(self.ESR_frequencies) > 1:
--- a/silq/parameters/acquisition_parameters.py
+++ b/silq/parameters/acquisition_parameters.py
@@ -1735,7 +1735,8 @@ class NMRParameter(AcquisitionParameter):
 
     """
     def __init__(self, name: str = 'NMR',
-                 names: List[str] = ['flips', 'flip_probability', 'up_proportions'],
+                 names: List[str] = ['flips', 'flip_probability', 'up_proportions',
+                                     'state_probability_0', 'state_probability_1'],
                  **kwargs):
         """
         Parameter used to determine the Rabi frequency
@@ -1759,7 +1760,8 @@ class NMRParameter(AcquisitionParameter):
         names = []
 
         for name in self._names:
-            if name in ['flips', 'flip_probability', 'up_proportions']:
+            if name in ['flips', 'flip_probability', 'up_proportions',
+                        'state_probability_0', 'state_probability_1']:
                 if len(self.ESR_frequencies) == 1:
                     names.append(name)
                 else:
@@ -1858,17 +1860,19 @@ class NMRParameter(AcquisitionParameter):
         self.ESR['ESR_pulses'] = updated_ESR_pulses
 
     def analyse(self, traces: Dict[str, Dict[str, np.ndarray]] = None):
-        """Analyse flipping events between nuclear states
+        """Analyse flipping events between nuclear states and determine nuclear state
 
         Returns:
             (Dict[str, Any]): Dict containing:
 
-            :results_read (dict): `analyse_traces` results for each read
+            * **results_read** (dict): `analyse_traces` results for each read
               trace
-            :up_proportions_{idx} (np.ndarray): Up proportions, the
+            * **up_proportions_{idx}** (np.ndarray): Up proportions, the
               dimensionality being equal to ``NMRParameter.samples``.
               ``{idx}`` is replaced with the zero-based ESR frequency index.
-            :Results from `analyse_flips`. These are:
+            * **state_probability_0_{idx}**, **state_probability_1_{idx}** (np.ndarray):
+              respectively up, down nuclear spin state proportions when reading-out ESR1 transition
+            * Results from `analyse_flips`. These are:
 
               - flips_{idx},
               - flip_probability_{idx}
@@ -1883,7 +1887,7 @@ class NMRParameter(AcquisitionParameter):
               each sample. The values that do not satisfy the filter are set to
               ``np.nan``.
 
-              :filtered_scans_{idx1}{idx2}:
+              * **filtered_scans_{idx1}{idx2}**:
         """
         if traces is None:
             traces = self.traces
@@ -1908,6 +1912,8 @@ class NMRParameter(AcquisitionParameter):
                                      self.ESR['shots_per_frequency'],
                                      points_per_shot))
         up_proportions = np.zeros((len(self.ESR_frequencies), self.samples))
+        state_probability_0 = np.zeros(len(self.ESR_frequencies))
+        state_probability_1 = np.zeros(len(self.ESR_frequencies))
         for f_idx, ESR_frequency in enumerate(self.ESR_frequencies):
             for sample in range(self.samples):
                 # Create array containing all read traces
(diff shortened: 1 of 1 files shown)


silq/parameters/acquisition_parameters.py
```diff
@@ -1858,17 +1860,19 @@ def ESR_frequencies(self, ESR_frequencies: List):
...
 
     def analyse(self, traces: Dict[str, Dict[str, np.ndarray]] = None):
-        """Analyse flipping events between nuclear states
+        """Analyse flipping events between nuclear states and determine nuclear state
 
         Returns:
             (Dict[str, Any]): Dict containing:
 
-            :results_read (dict): `analyse_traces` results for each read
+            * **results_read** (dict): `analyse_traces` results for each read
```
maij, line 1868: Is there a reason you're reformatting the docstrings like this?

silq/parameters/acquisition_parameters.py
```diff
@@ -1858,17 +1860,19 @@ def ESR_frequencies(self, ESR_frequencies: List):
...
               trace
-            :up_proportions_{idx} (np.ndarray): Up proportions, the
+            * **up_proportions_{idx}** (np.ndarray): Up proportions, the
               dimensionality being equal to ``NMRParameter.samples``.
               ``{idx}`` is replaced with the zero-based ESR frequency index.
-            :Results from `analyse_flips`. These are:
+            * **state_probability_0_{idx}**, **state_probability_1_{idx}** (np.ndarray):
+              respectively up, down nuclear spin state proportions when reading-out ESR1 transition
```
maij, line 1874: This isn't generic, ESR1 shouldn't be mentioned here because the way the NMR parameter works is to measure an up-proportion for arbitrary ESR pulses.

For example, one might measure with ESR_frequencies:
`ESR_frequencies = [[ESR2, ESR1], ESR1]`
or
`ESR_frequencies = [ESR1, ESR2, ESR1]`
or
`ESR_frequencies = [ESR1, ESR2, ESR3]`

The first example shows a result from 2 ESR stimuli, one which addressed both ESR1 + ESR2 and the other which addresses ESR1. The first does not give you a simple 'state_probability'!
The last example shows that you can measure more than 2 ESR frequencies, e.g. for antimony this is inevitable.

So I have 2 suggestions:
- Make `state_probability_0_{idx}` more generic, and matched with the number of ESR_frequencies
I would suggest renaming this to something like `success_probability` which inherits its name from standard binomial distribution nomenclature. You can also include 'failure_probability`, but this is redundant (if you do, make sure you have one threshold be inclusive! i.e. use >= or <= ).
In this way `success_probability` would simply replace `state_probability_1`.
- Include ESR_frequencies in the dataset as a setpoint, if it isn't already.

silq/parameters/acquisition_parameters.py
```diff
@@ -1928,10 +1934,19 @@ def analyse(self, traces: Dict[str, Dict[str, np.ndarray]] = None):
                 up_proportions[f_idx, sample] = read_result['up_proportion']
                 results['results_read'].append(read_result)
 
+            state_probability_0[f_idx] = np.sum(up_proportions[f_idx] <
+                                                self.threshold_up_proportion)/self.samples
+            state_probability_1[f_idx] = np.sum(up_proportions[f_idx] >
+                                                self.threshold_up_proportion)/self.samples
```
maij, line 1940:
```suggestion
            success_probability[f_idx] = np.mean(up_proportions[f_idx] > self.threshold_up_proportion)
```

silq/parameters/acquisition_parameters.py
```diff
@@ -1928,10 +1934,19 @@ def analyse(self, traces: Dict[str, Dict[str, np.ndarray]] = None):
...
+
             if len(self.ESR_frequencies) > 1:
                 results[f'up_proportions_{f_idx}'] = up_proportions[f_idx]
+                results[f'state_probability_0_{f_idx}'] = state_probability_0[f_idx]
+                results[f'state_probability_1_{f_idx}'] = state_probability_1[f_idx]
             else:
                 results['up_proportions'] = up_proportions[f_idx]
+                results['state_probability_0'] = state_probability_0[f_idx]
+                results['state_probability_1'] = state_probability_1[f_idx]
```
maij, line 1949: Here I think we can just store `state_probability` or `success_probability` directly to the dataset without splitting it out based on `f_idx`, it will just be a 2D array. I believe Serwan has already implemented something like this for the new measurement parameters.

We can check in on that later.

silq/parameters/acquisition_parameters.py
RostyslavSavytskyy: When I was using the same formatting style that it already had, Serwan said I should use different one, which I found elsewhere, so now I have no idea which style to use...

silq/parameters/acquisition_parameters.py
RostyslavSavytskyy: But isn't it worth having this similar to flip_probability? I.e. when we readout separately different ESR peaks, we calculate flip probability and store this separately. The same I would guess to be for this "success_probability"

silq/parameters/acquisition_parameters.py
RostyslavSavytskyy: "Include ESR_frequencies in the dataset as a setpoint, if it isn't already." - do you mean by this that I should add saving of different ESR frequencies that we are using for readout?

maij: Sorry, I realise that was beyond the scope of this PR, so I take back that suggestion, I will deal with that in a separate PR in the future.

RostyslavSavytskyy: ok, cool

RostyslavSavytskyy:
Ok, so I applied above requests except I still left splitting of success_probability based on f_idx.

silq/parameters/acquisition_parameters.py
```diff
@@ -1858,17 +1860,19 @@ def ESR_frequencies(self, ESR_frequencies: List):
         self.ESR['ESR_pulses'] = updated_ESR_pulses
 
     def analyse(self, traces: Dict[str, Dict[str, np.ndarray]] = None):
-        """Analyse flipping events between nuclear states
+        """Analyse flipping events between nuclear states and determine nuclear state
```
maij, line 1863: This is more in line with what the code is doing.
```suggestion
        """Analyse flipping events between nuclear states and determine the success rate of each ESR pulse (i.e. samples exceeding self.threshold_up_proportion).
```

nulinspiratie:
I've indeed also implemented this functionality for the new acquisition parameters, but I understand that this PR is necessary.

It's a good improvement, but I don't like the term `success_probability`, as it's not clear when something is a success. In my other PR I've used `state_probability`, I think that works better.
In the other PR I've also added `state`, similar to `flips`.

And I don't think the state probabilities should be combined into an array. The reason is that in the new acquisition parameter you can add labels to each of the states, and this doesn't work with arrays

maij:
> I've indeed also implemented this functionality for the new acquisition parameters, but I understand that this PR is necessary.
> 
> It's a good improvement, but I don't like the term `success_probability`, as it's not clear when something is a success. In my other PR I've used `state_probability`, I think that works better.
> In the other PR I've also added `state`, similar to `flips`.
> 
> And I don't think the state probabilities should be combined into an array. The reason is that in the new acquisition parameter you can add labels to each of the states, and this doesn't work with arrays

The main reason I'm against `state_probability` is because what you're really looking for is whether a certain set of 'ESR pulses' is actually resonant, which might correspond to a specific state, but it might not. 

I gave the example earlier:
`ESR_frequencies = [[ESR1], [ESR1, ESR2]]`
The second pulse doesn't indicate a single state, but pulses 2 underlying states.

So that's why I was suggesting something like `ESR_success_rate` in the end, because it avoids making an assumption about what the ESR result means, as it fundamentally acts to distinguish successful ESR from background dark counts.

Anyway, I won't contest this any further, just wanted to make sure we're on the same page.

nulinspiratie:
Ah, I get your point, and it's well taken. In fact, I have run into this a few times already, as to distinguish the control donor nuclear state, two or more ESR pulses need to be applied since we have multiple possible resonance lines. However, so far we could always still find a proper nuclear state label (e.g. `control_down`), even though it corresponds to multiple ESR pulses.

I agree that there will be situations where there isn't a proper state label, but I think this will be fairly uncommon. What do you think?

RostyslavSavytskyy:
I agree with both of you, but maybe more with Serwan concerning the naming - success probability - might be too generic and confusing for those who didn't look closely at the code. But I am fine with both - your final call and I will modify appropriately.

maij:
I haven't used this new labelling feature, but that does sound like a reasonable solution to the problem.
I'll retract my review now.

RostyslavSavytskyy:
So I renamed back to state_probability as requested. Ready to be merged.

maij:
@nulinspiratie Have your changes been fully complete?

RostyslavSavytskyy:
I think so, I renamed as agreed and the rest should be fine

RostyslavSavytskyy:
Haha, thanks!

PR #202, opened 2020-04-17
added new EDSR parameter based on NMR parameter

RostyslavSavytskyy:
In the new NMR pulse sequence, there is an additional "read" post-pulse after NMR pulse that can be used for electron-readout in EDSR measurements. This parameter was not tested yet and requires update to the new pulse sequence first to work

### DIFF
--- a/silq/parameters/acquisition_parameters.py
+++ b/silq/parameters/acquisition_parameters.py
@@ -1959,8 +1959,7 @@ class EDSRParameter(NMRParameter):
 
     def __init__(self, name: str = 'EDSR',
                  names: List[str] = ['flips', 'flip_probability', 'up_proportions',
-                                     'state_probability_0', 'state_probability_1',
-                                     'EDSR_up_proportion'],
+                                     'success_probability', 'EDSR_up_proportion'],
                  **kwargs):
         super().__init__(name=name,
                          names=names,
--- a/silq/parameters/acquisition_parameters.py
+++ b/silq/parameters/acquisition_parameters.py
@@ -2009,21 +2009,13 @@ class EDSRParameter(NMRParameter):
         """
         results = super().analyse(traces)
 
-        # Extract points for read DC pulse after EDSR pulse
-        EDSR_read_traces_name = f"{self.NMR['post_pulse'].name}"
-        EDSR_read_traces = traces[EDSR_read_traces_name]['output']
-        EDSR_trace_points = EDSR_read_traces.shape[1]
-
-        self.EDSR_traces = np.zeros((self.samples, EDSR_trace_points))
-        self.EDSR_traces = EDSR_read_traces
         EDSR_read_result = analysis.analyse_traces(
-            traces=EDSR_read_traces,
+            traces=traces[self.NMR['post_pulse'].name]['output'],
             sample_rate=self.sample_rate,
             t_read=self.t_read,
             t_skip=self.t_skip,
             threshold_voltage=self.threshold_up_proportion)
-        EDSR_up_proportion = EDSR_read_result['up_proportion']
-        results['EDSR_up_proportion'] = EDSR_up_proportion
+        results['EDSR_up_proportion'] = EDSR_read_result['up_proportion']
 
         return results
 
--- a/silq/parameters/acquisition_parameters.py
+++ b/silq/parameters/acquisition_parameters.py
@@ -23,7 +23,7 @@ from silq.tools.general_tools import SettingsClass, clear_single_settings, \
 
 __all__ = ['AcquisitionParameter', 'DCParameter', 'TraceParameter',
            'DCSweepParameter', 'EPRParameter', 'ESRParameter',
-           'NMRParameter', 'VariableReadParameter', 'BlipsParameter',
+           'NMRParameter', 'EDSRParameter', 'VariableReadParameter', 'BlipsParameter',
            'FlipNucleusParameter', 'FlipFlopParameter', 'NeuralNetworkParameter',
            'NeuralRetuneParameter','ESRRamseyDetuningParameter']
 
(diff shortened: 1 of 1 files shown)


maij:
PR #201 needs to be merged before this one, so I'll delay review until that time.

nulinspiratie:
Hmm at the moment I'm not sure if it's worthwhile adding this to the master branch, given its overlap with the new pulse sequences. I suggest somewhere this week we have a discussion to see if it makes sense to switch to the new pulse sequences, and we can decide based on that

RostyslavSavytskyy:
What kind of overlap is there? It's completely different parameter... It would be good to add this as well so that we had a clean merging of flip-flop branch with master branch during the switch to the new SilQ. Otherwise some stuff will not be working. Alternatively, I drag this myself separately in flip-flop branch if you prefer, but it's a bit annoying...

RostyslavSavytskyy:
If you mean that the new SilQ does not need this - there is the same argument for many other parameters that eventually should be deleted, but I don't see the reason why when deleting those other parameters at some point later, we delete this one with them as well?

maij:
@RostyslavSavytskyy @nulinspiratie Since you're already talking about it...

Can we put together a document (maybe an issue?) that details what will be done in 'old' silq and what will be done in 'new' silq? I think this should become a somewhat standard procedure now that we're thinking of bigger milestones and releases.

RostyslavSavytskyy:
I updated this branch to the master level, i.e. now it should only add EDSRParameter to the acquisition parameters, which I think should be safe and have no overlaps with other branches. If this parameter is not used in the future, it can easily be deleted with any other unnecessary parameters. So, please have a look at this pull request as well (should be quick) and we can add this to the master branch so that we did not have to drag this anyhow with the flip-flop branch. Let me know of any issues.

silq/parameters/acquisition_parameters.py
```diff
@@ -1965,6 +1965,69 @@ def analyse(self, traces: Dict[str, Dict[str, np.ndarray]] = None):
...
+
+        Returns:
+            (Dict[str, Any]): Dict containing:
+            * all results from NMRParameter
+            * **EDSR_up_proportion**: electron spin-up proportion right after EDSR (NMR) pulse.
+        """
+        results = super().analyse(traces)
+
+        # Extract points for read DC pulse after EDSR pulse
+        EDSR_read_traces_name = f"{self.NMR['post_pulse'].name}"
```
nulinspiratie, line 2013: This does not need to be converted to a string, shouldn't it already be a string? If not, then simply use str(self.NMR['post_pulse'].name)

nulinspiratie: Also, how is `post_pulse` defined? There's no pulse added to the class, so won't this by default raise an error?

nulinspiratie: And isn't this variable used just once? I don't see the reason of creating a separate variable for this

silq/parameters/acquisition_parameters.py
```diff
@@ -1965,6 +1965,69 @@ def analyse(self, traces: Dict[str, Dict[str, np.ndarray]] = None):
...
+        EDSR_read_traces = traces[EDSR_read_traces_name]['output']
+        EDSR_trace_points = EDSR_read_traces.shape[1]
+
+        self.EDSR_traces = np.zeros((self.samples, EDSR_trace_points))
```
nulinspiratie, line 2017: This line is unnecessary

silq/parameters/acquisition_parameters.py
```diff
@@ -1965,6 +1965,69 @@ def analyse(self, traces: Dict[str, Dict[str, np.ndarray]] = None):
...
+        self.EDSR_traces = EDSR_read_traces
```
nulinspiratie, line 2018: This line is also unnecessary. I don't think it's needed to have an additional `EDSR_traces` attribute since it can already be accessed via self.traces. Otherwise it would also make sense to add `ESR_traces` and `NMR_traces`

silq/parameters/acquisition_parameters.py
```diff
@@ -1965,6 +1965,69 @@ def analyse(self, traces: Dict[str, Dict[str, np.ndarray]] = None):
...
+        EDSR_read_result = analysis.analyse_traces(
+            traces=EDSR_read_traces,
+            sample_rate=self.sample_rate,
+            t_read=self.t_read,
+            t_skip=self.t_skip,
+            threshold_voltage=self.threshold_up_proportion)
+        EDSR_up_proportion = EDSR_read_result['up_proportion']
```
nulinspiratie, line 2025: This line is also unnecessary, can be combined with line below

silq/parameters/acquisition_parameters.py
RostyslavSavytskyy: post_pulse is defined by default in the NMRPulseSequence to be DCPulse('read', acquire=True). So there shouldn't be errors.
Otherwise, I've already corrected your suggestions.

silq/parameters/acquisition_parameters.py
RostyslavSavytskyy: corrected.

silq/parameters/acquisition_parameters.py
RostyslavSavytskyy: corrected.

silq/parameters/acquisition_parameters.py
RostyslavSavytskyy: corrected.

RostyslavSavytskyy:
Great, yes i think we should merge for now to smooth our transition to the new Silq so that we could easily use our old stuff until we complete the switch.

PR #203, opened 2020-04-17
added adaptive threshold up proportion function in analysis

RostyslavSavytskyy:
For now it is only function not used anywhere yet. Can implement it later into analyse_flips function and NMR_parameter, etc. But can be already tested as it is


silq/analysis/analysis.py
```diff
@@ -715,6 +717,38 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+        if (len(proportion_space) - peak_idxs[0]) / len(proportion_space) > 0.5:
+            true_slice = slice(*[peak_idxs[0], len(proportion_space)])
+            true_idx = np.argmin(np.round(gaussian_up_proportions[true_slice], 4)) + \
+                       peak_idxs[0]
+        else:
+            true_slice = slice(*[0, peak_idxs[0]])
+            true_idx = np.argmin(np.round(gaussian_up_proportions[true_slice], 4))
+    else:
+        true_slice = slice(*[min(peak_idxs), max(peak_idxs)])
+        true_idx = np.argmin(np.round(gaussian_up_proportions[true_slice], 4)) + \
```
maij, line 744: Could you explain the rounding please?
Is it necessary?

silq/analysis/analysis.py
```diff
@@ -715,6 +717,38 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+                     f'found, using threshold 0.5')
+        true_idx = round((len(proportion_space) - 1) / 2)
+    elif len(peak_idxs) == 1:
+        if (len(proportion_space) - peak_idxs[0]) / len(proportion_space) > 0.5:
+            true_slice = slice(*[peak_idxs[0], len(proportion_space)])
+            true_idx = np.argmin(np.round(gaussian_up_proportions[true_slice], 4)) + \
+                       peak_idxs[0]
+        else:
+            true_slice = slice(*[0, peak_idxs[0]])
+            true_idx = np.argmin(np.round(gaussian_up_proportions[true_slice], 4))
```
maij, line 741: I don't think any of this is necessary, if you've found only 1 peak then it doesn't matter where you place the threshold, it could be either at 0 or 1, depending on whether the peak is in the left- or right-half of the proportion space.

Do you have a counter-example?

RostyslavSavytskyy:
Thanks Mark for the questions/suggestions, I will come back for them. At the moment I had to merge with other branches, where this adaptive threshold will be used. So want to raise a note that this branch should be pulled the last, i.e. after state_probability in NMR parameter and EDSR parameter branches.

silq/analysis/analysis.py
```diff
@@ -715,6 +717,38 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+def analyse_threshold_up_proportion(up_proportions_arrs: np.ndarray,
+                                    shots_per_frequency: int):
+    proportion_space = np.linspace(0, 1, num=shots_per_frequency + 1,
+                                   endpoint=True)
+    up_proportions_arrs = up_proportions_arrs.reshape(1, -1)
+    kernel = gaussian_kde(up_proportions_arrs)
+    gaussian_up_proportions = kernel(proportion_space)
+    peak_idxs = peakutils.peak.indexes(gaussian_up_proportions,
+                                       thres=0.5/up_proportions_arrs.shape[-1],
+                                       min_dist=shots_per_frequency/5)
```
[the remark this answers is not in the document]
RostyslavSavytskyy: Generally, I agree, but I saw the situations when the spin-up proportion is low (spread from 0.1 to 0.4) and there was a "dip" (missing shots) somewhere around 0.25. In this case according to the code it will define threshold as 0.25, which actually should be 0.5 or so... So I would be careful with this...

silq/analysis/analysis.py
```diff
@@ -715,6 +717,38 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+    if len(peak_idxs) == 0:
+        logger.debug(f'Adaptive thresholding routine: 0 peaks were '
+                     f'found, using threshold 0.5')
+        true_idx = round((len(proportion_space) - 1) / 2)
```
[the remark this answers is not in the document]
RostyslavSavytskyy: agree

silq/analysis/analysis.py
```diff
@@ -715,6 +717,38 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+                       peak_idxs[0]
+        else:
+            true_slice = slice(*[0, peak_idxs[0]])
+            true_idx = np.argmin(np.round(gaussian_up_proportions[true_slice], 4))
+    else:
+        true_slice = slice(*[min(peak_idxs), max(peak_idxs)])
+        true_idx = np.argmin(np.round(gaussian_up_proportions[true_slice], 4)) + \
+                   peak_idxs[0]
+
+    threshold_up_proportion = proportion_space[true_idx]
```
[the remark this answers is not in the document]
RostyslavSavytskyy: sorry for my English, was a bit confused, all good now

silq/analysis/analysis.py
```diff
@@ -715,6 +717,38 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+    elif len(peak_idxs) == 1:
+        if (len(proportion_space) - peak_idxs[0]) / len(proportion_space) > 0.5:
+            true_slice = slice(*[peak_idxs[0], len(proportion_space)])
+            true_idx = np.argmin(np.round(gaussian_up_proportions[true_slice], 4)) + \
+                       peak_idxs[0]
+        else:
+            true_slice = slice(*[0, peak_idxs[0]])
+            true_idx = np.argmin(np.round(gaussian_up_proportions[true_slice], 4))
+    else:
+        true_slice = slice(*[min(peak_idxs), max(peak_idxs)])
```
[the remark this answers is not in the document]
RostyslavSavytskyy: image

slice function doesn't accept more than 3 arguments, so if we have multiple peaks, the above version will fail

silq/analysis/analysis.py
RostyslavSavytskyy: Since the minimum value of gaussian distribution sometimes gets to 1e-34 at spin-up proportion of 1 or 0 - without rounding we will get threshold = 1 or 0. But the idea is to potentially use the threshold as a function of charge stability (any jumps in spin-up proportion can be detected) throughout our long measurements. Thus, when we don't have two gaussian peaks (but just one when nucleus is not flipping), we will have weird 0 or 1 threshold. To eliminate these jumps by rounding to the 4 digit after point, will make threshold 0.6 instead of 1 for example. It is not elegant solution, but I haven't found yet a better one.

silq/analysis/analysis.py
RostyslavSavytskyy: This refers to the same issue above, I want to have threshold oscillating somewhere around 0.5 (in the middle of spin-up proportion space) so that it would have potential application in the future.

silq/analysis/analysis.py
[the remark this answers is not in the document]
maij: In that case, it sounds like the smoothing routine didn't work correctly. It should be able to smooth out missing datapoints adequately.

Just putting some of my stats thoughts down here for my future reference:
The standard deviation of a binomial distribution under the CLT is `sigma/sqrt(n)` where `n = samples` and `sigma` is the intrinsic standard deviation of the binomial distribution: `p(1-p)` where `p` is the centre of the peak, e.g. 0.2 or 0.8. The largest value for `sigma` is when `p = 0.5`, so `sigma = 0.25`. I think a conservative choice for the gaussian_kde bandwidth should be `0.25/sqrt(n)`, taking care to note that n is NOT the `shots_per_frequency` but the number of samples gathered for the up-proportions of this round.
We could also think about introducing a confidence level on the threshold found, which would have to be determined somehow. This could allow the use of past threshold with their confidences as weights to be used to inform the current threshold. This is particularly useful if you're using very few samples.

maij:
Just putting some of my stats thoughts down here for my future reference:
The standard deviation of a binomial distribution under the CLT is `sigma/sqrt(n)` where `n = samples` and `sigma` is the intrinsic standard deviation of the binomial distribution: `p(1-p)` where `p` is the centre of the peak, e.g. 0.2 or 0.8. The largest value for `sigma` is when `p = 0.5`, so `sigma = 0.25`. I think a conservative choice for the gaussian_kde bandwidth should be `0.25/sqrt(n)`, taking care to note that n is NOT the `shots_per_frequency` but the number of samples gathered for the up-proportions of this round.
We could also think about introducing a confidence level on the threshold found, which would have to be determined somehow. This could allow the use of past threshold with their confidences as weights to be used to inform the current threshold. This is particularly useful if you're using very few samples.

silq/analysis/analysis.py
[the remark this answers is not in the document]
maij: Right, I understand that but figured you should never see more than 2 peaks. This seems to be a foundational issue that this technique can sometimes return more than 2 peaks. As I understood, the `scipy.stats.gaussian_kde` function applied to a bimodal distribution should *oversmooth* the data, as opposed to undersmoothing resulting in false peaks.

Can you comment on this at all? In your experience, with real data, do you often find the data is undersmoothed? When I first implemented my version of the 'dumb' gaussian KDE with an arbitrary bandwidth I never saw more than 2 peaks.

silq/analysis/analysis.py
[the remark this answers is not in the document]
RostyslavSavytskyy: if there is a dip - there is dip, and it will be present in density distribution, I am not sure this can be smoothed much - just the depth will change, but the dip will be there...

silq/analysis/analysis.py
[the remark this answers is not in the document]
RostyslavSavytskyy: From what I have seen - more than 2 peaks might happen in about 1% of the cases, maybe less, so just in case I made it for multiple peaks.

RostyslavSavytskyy:
Yes, agree - introducing previous values for threshold and adjusting respectively the current value will be beneficial

silq/analysis/analysis.py
```diff
@@ -715,8 +717,44 @@ def analyse_EPR(empty_traces: np.ndarray,
             'mean_high_blip_duration': results_read['mean_high_blip_duration']}
 
 
-def analyse_flips(up_proportions_arrs: List[np.ndarray],
-                  threshold_up_proportion: Union[Sequence, float]):
+def analyse_threshold_up_proportion(up_proportions_arrs: np.ndarray,
```
nulinspiratie, line 720: Am I correct in thinking that this adaptive threshold up proportion is only for the case where there is a single up_proportion array? If so, this should be written in the function docstring, and maybe the function name should also reflect this.  `analyse_threshold_up_proportion` should be replaced with `determine_threshold_up_proportion`, since you're analysing traces to determine the threshold.

silq/analysis/analysis.py
```diff
@@ -715,8 +717,44 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+                                    shots_per_frequency: int):
+    if shots_per_frequency < 25:
+        num_shots = 25   # to make sure algorithm works
```
nulinspiratie, line 723: This comment is quite unclear, why 25? Is it that it otherwise has too few points to converge? The reason should be clearly explained in a comment, especially because it seems like an arbitrary value.
Also these four lines can be replaced by: `num_shots = max(shots_per_frequency, 25)

silq/analysis/analysis.py
```diff
@@ -715,8 +717,44 @@ def analyse_EPR(empty_traces: np.ndarray,
             'mean_high_blip_duration': results_read['mean_high_blip_duration']}
 
 
-def analyse_flips(up_proportions_arrs: List[np.ndarray],
-                  threshold_up_proportion: Union[Sequence, float]):
+def analyse_threshold_up_proportion(up_proportions_arrs: np.ndarray,
+                                    shots_per_frequency: int):
```
nulinspiratie, line 721: Missing function docstring. I think it's especially needed in this case because the algorithm contains several steps and is note trivial

silq/analysis/analysis.py
```diff
@@ -715,8 +717,44 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+    if shots_per_frequency < 25:
+        num_shots = 25   # to make sure algorithm works
+    else:
+        num_shots = shots_per_frequency
+    proportion_space = np.linspace(0, 1, num=num_shots + 1, endpoint=True)
+    up_proportions_arrs = up_proportions_arrs.reshape(1, -1)
+    kernel = gaussian_kde(up_proportions_arrs)
+    gaussian_up_proportions = kernel(proportion_space)
+    peak_idxs = peakutils.peak.indexes(gaussian_up_proportions,
```
nulinspiratie, line 730: What happens if there are only a few shots? Does this method still work or does it break down?

silq/analysis/analysis.py
```diff
@@ -715,8 +717,44 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+                                       thres=0.5/up_proportions_arrs.shape[-1],
+                                       min_dist=num_shots/5)
+    if len(peak_idxs) == 0:
+        logger.debug(f'Adaptive thresholding routine: 0 peaks were '
+                     f'found, using threshold 0.5')
+        trough_idx = (num_shots + 1) // 2
+    elif len(peak_idxs) == 1:
```
nulinspiratie, line 737: I can't easily understand what's happening in this section of code. It needs some comments, and preferably also one or two intermediate variables. For isntance, is there a variable name you could give to: `np.round(gaussian_up_proportions[trough_slice], 4)`?

RostyslavSavytskyy:
Implemented what you requested. Should be ready to go.

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+    # If only one peak is detected (nuclear spin did not flip), depending on whether
+    # it is in the lower (<0.5) or the upper (>=0.5) half of proportion space,
+    # the minimum index of density distribution is found respectively above or below
+    # this peak's index.
+    # Since the density distribution values reach 1e-20 at the edges of the
+    # proportion space (i.e. at 0 or 1), to smooth a bit the threshold jumps between
+    # the measurement points, rounding to the 4th digit is implemented.
+    elif len(peak_idxs) == 1:
+        if peak_idxs[0] < middle_point:
+            trough_slice = slice(*[peak_idxs[0], num_shots + 1])
```
nulinspiratie, line 772: doesnt need to be a list: `slice(peak_idxs[0], num_shots + 1)`

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+            search_space = np.round(gaussian_up_proportions[trough_slice], 4)
+            trough_idx = peak_idxs[0] + np.argmin(search_space)
```
nulinspiratie, line 774: I'm not following these lines, it needs a bit more commenting. Are you basically finding the first up proportion above the peak and setting that as your threshold? I'm not seeing why that is the right approach, can you explain?

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+        else:
+            trough_slice = slice(*[0, peak_idxs[0]])
```
nulinspiratie, line 776: Same change as above

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+            search_space = np.round(gaussian_up_proportions[trough_slice], 4)
+            trough_idx = np.argmin(search_space)
```
nulinspiratie, line 778: Why are you also taking the min here?

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+    # If more than one peak is detected, then minimum is found between the lowest and the highest peak.
+    else:
+        trough_slice = slice(*[min(peak_idxs), max(peak_idxs)])
+        search_space = np.round(gaussian_up_proportions[trough_slice], 4)
+        trough_idx = peak_idxs[0] + np.argmin(search_space)
```
nulinspiratie, line 783: Why the min? Why not the average?

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
             'mean_high_blip_duration': results_read['mean_high_blip_duration']}
 
 
-def analyse_flips(up_proportions_arrs: List[np.ndarray],
-                  threshold_up_proportion: Union[Sequence, float]):
+def determine_threshold_up_proportion(up_proportions_arr: np.ndarray,
```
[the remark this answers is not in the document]
nulinspiratie: I've also given it some thought, and I think it makes sense to change the function name such that it reflects that this is only when a single state is measured. Because there's also another function (in another PR) that determines the threshold up proportion when multiple arrays are provided.

Maybe something like `determine_threshold_up_proportion_single_state`? If you have another suggestion let me know

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+            `analyse_traces`. Up proportion is the proportion of traces that
+            contain blips. The length is number of samples.
+        shots_per_frequency: Integer number of ESR shots/traces, which are
+            used to determine the up proportion.
+
+    Returns:
+        (float) threshold up-proportion
+    """
+
+    num_shots = max(shots_per_frequency, 25)
```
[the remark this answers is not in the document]
RostyslavSavytskyy: corrected

silq/analysis/analysis.py
maij: The algorithm aims to find the minimum overlap point between 2 gaussian distributions, hence taking `np.argmin()` of the kernel function, rather than the average of the 2 peaks. 
If the 2 peaks have different widths then the average is not the best choice, however in this application of the algorithm all data is smoothed to the same width I believe. There is room for improvement here.

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+
+    proportion_space = np.linspace(0, 1, num=num_shots + 1, endpoint=True)
+    up_proportions_arr = up_proportions_arr.reshape(1, -1)    # don't think it's needed
+    kernel = gaussian_kde(up_proportions_arr)
+    gaussian_up_proportions = kernel(proportion_space)
+
+    # Finding indexes of density distribution peaks. Might still require some fine adjustments in
+    # 'thres' and 'min_dist' parameters.
+    peak_idxs = peakutils.peak.indexes(gaussian_up_proportions,
+                                       thres=0.5/up_proportions_arr.shape[-1],
```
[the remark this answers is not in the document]
RostyslavSavytskyy: corrected

silq/analysis/analysis.py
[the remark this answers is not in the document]
RostyslavSavytskyy: corrected.

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+        (float) threshold up-proportion
+    """
+
+    num_shots = max(shots_per_frequency, 25)
+    # Algorithm fails if the number of shots is less than 25 (not exact number, can still be adjusted).
+    # The reason for this is that in the following division of up_proportion space by the number of shots
+    # there are too few points to perform kernel density estimation.
+
+    proportion_space = np.linspace(0, 1, num=num_shots + 1, endpoint=True)
+    up_proportions_arr = up_proportions_arr.reshape(1, -1)    # don't think it's needed
```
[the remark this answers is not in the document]
RostyslavSavytskyy: yes, already corrected.

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+    peak_idxs = peakutils.peak.indexes(gaussian_up_proportions,
+                                       thres=0.5/up_proportions_arr.shape[-1],
+                                       min_dist=num_shots/5)
+    # If no peaks are detected (something failed), by default we set threshold to 0.5.
+    middle_point = (num_shots + 1) // 2
+    if len(peak_idxs) == 0:
+        logger.debug(f'Adaptive thresholding routine: 0 peaks were '
+                     f'found, using threshold 0.5')
+        trough_idx = middle_point
+    # If only one peak is detected (nuclear spin did not flip), depending on whether
```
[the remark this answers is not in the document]
RostyslavSavytskyy: so, it's already ok, right?

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+
+    num_shots = max(shots_per_frequency, 25)
+    # Algorithm fails if the number of shots is less than 25 (not exact number, can still be adjusted).
+    # The reason for this is that in the following division of up_proportion space by the number of shots
+    # there are too few points to perform kernel density estimation.
+
+    proportion_space = np.linspace(0, 1, num=num_shots + 1, endpoint=True)
+    up_proportions_arr = up_proportions_arr.reshape(1, -1)    # don't think it's needed
+    kernel = gaussian_kde(up_proportions_arr)
+    gaussian_up_proportions = kernel(proportion_space)
```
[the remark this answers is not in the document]
RostyslavSavytskyy: corrected.

silq/analysis/analysis.py
RostyslavSavytskyy: corrected.

silq/analysis/analysis.py
RostyslavSavytskyy: corrected.

silq/analysis/analysis.py
RostyslavSavytskyy: I agree with Mark

silq/analysis/analysis.py
RostyslavSavytskyy: No, if there is only one peak, it finds its index, then it looks for the minimum of density distribution (somewhere far from the Gaussian peak) in the proportion space between the peak maximum and the rest of proportion space. There is a condition that it searches for density minimum in proportion space that is larger. E.g. if peak is at 0.2 proportion space, minimum will be found in the range between 0.2 and 1. If peak was at 0.6, the minimum would be found somewhere between 0 and 0.6.

RostyslavSavytskyy: I added these comments below. Should I add more?
# If only one peak is detected (nuclear spin did not flip), depending on whether
    # it is in the lower (<0.5) or the upper (>=0.5) half of proportion space,
    # the minimum index of density distribution is found respectively above or below
    # this peak's index.
    # Since the density distribution values reach 1e-20 at the edges of the
    # proportion space (i.e. at 0 or 1), to smooth a bit the threshold jumps between
    # the measurement points, rounding to the 4th digit is implemented.

silq/analysis/analysis.py
RostyslavSavytskyy: the same idea as is answered above and below

silq/analysis/analysis.py
```diff
@@ -715,8 +717,81 @@ def analyse_EPR(empty_traces: np.ndarray,
             'mean_high_blip_duration': results_read['mean_high_blip_duration']}
 
 
-def analyse_flips(up_proportions_arrs: List[np.ndarray],
-                  threshold_up_proportion: Union[Sequence, float]):
+def determine_threshold_up_proportion(up_proportions_arr: np.ndarray,
+                                      shots_per_frequency: int):
+    """ Determine threshold up-proportion for nuclear readout
```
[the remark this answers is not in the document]
RostyslavSavytskyy: corrected

silq/analysis/analysis.py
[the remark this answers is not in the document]
nulinspiratie: Isn't the warning now always shown? I believe you need a check.

silq/analysis/analysis.py
```diff
@@ -715,8 +717,85 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+        shots_per_frequency: Integer number of ESR shots/traces, which are
+            used to determine the up proportion.
+
+    Returns:
+        (float) threshold up-proportion
+    """
+    assert isinstance(up_proportions_arr, np.ndarray), 'Up proportions must be a 1D array, not a list.'
+
+    num_shots = max(shots_per_frequency, 25)
+    logger.warning("Number of shots is < 25, 25 shots is used for threshold analysis.")
```
nulinspiratie, line 745: Warning is a bit vague atm. What about:
`logger.warning(f'Readout shots {shots_per_frequency} < 25. This may cause issues when finding an optimal threshold up proportion.')

silq/analysis/analysis.py
```diff
@@ -715,8 +717,85 @@ def analyse_EPR(empty_traces: np.ndarray,
...
+    # Algorithm fails if the number of shots is less than 25 (not exact number, can still be adjusted).
```
nulinspiratie, line 746: Good comment, but it would be good adding this to the docstring, so at the top of the function. The reason is that this is part of the behaviour of the function, so it should be mentioned above

silq/analysis/analysis.py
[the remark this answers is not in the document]
nulinspiratie: The comments are good, but by docstring I mean the piece of text at the top of the function. This is generally the information that the user will look at, not the comments in the code. The docstring should therefore contain a summarized version of the function behaviour

silq/analysis/analysis.py
nulinspiratie: yep I think it's good

silq/analysis/analysis.py
nulinspiratie: Hmm maybe add it as a `# TODO {text}` comment?

silq/analysis/analysis.py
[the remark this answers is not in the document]
RostyslavSavytskyy: oops, corrected

silq/analysis/analysis.py
RostyslavSavytskyy: corrected.

silq/analysis/analysis.py
RostyslavSavytskyy: corrected.

silq/analysis/analysis.py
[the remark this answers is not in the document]
RostyslavSavytskyy: corrected.

silq/analysis/analysis.py
RostyslavSavytskyy: corrected.

RostyslavSavytskyy:
I don't think EDSRParameter is necessary here, so should be good to merge without it, but it is needed overall as well.

nulinspiratie:
> I don't think EDSRParameter is necessary here, so should be good to merge without it, but it is needed overall as well.

Okay in that case can you remove it from the changed files?

RostyslavSavytskyy:
Ok, I removed EDSRParameter and added small corrections after testing on Tallulah. Should be now good to go.

RostyslavSavytskyy:
Are you still reviewing anything? Should I make more changes?

nulinspiratie:
I should have time to look at it today

--

Dr. Serwan Asaad

Research associate

Fundamental Quantum Technologies Laboratory

University of New South Wales
Sydney, Australia

s.asaad@unsw.edu.au

fqt.unsw.edu.au <http://www.fqt.unsw.edu.au/>

RostyslavSavytskyy:
Let me know if you have any issues with my additions when merging other branches, can clarify/actively modify a few names, etc.

PR #205, opened 2020-04-18
Make the path OS independent for get_pulse_sequence.

maij:
Small fix for unix-like paths.
I'm still unsure about [one line ](https://github.com/nulinspiratie/SilQ/blob/f70ec5687c08266e47a354f45b311117a86c6716/silq/__init__.py#L463), I can't figure out why pulse sequences have a '\\\\' in them by default. I can't find where this gets added, but it doesn't seem to be an issue for now so long as pulse sequences aren't created on a unix-like system, I suppose.


nulinspiratie:
I think at some point we should make the conversion to pathlib, it circumvents any such issues

PR #207, opened 2020-04-22
fixed IQ modulation mode for single frequency pulses

RostyslavSavytskyy:
In IQ modulation mode if you pulse only with single frequency, i.e. you don't need IQ modulation, it shows an error. It is fixed now - was copied by Serwan from Queenie.


RostyslavSavytskyy:
Reverted MultiSinePulse additions from this pull request, good to go for now, unless you see some additional changes to be made

PR #208, opened 2020-04-22
added post_pulse after NMR pulse

RostyslavSavytskyy:
added post_pulse after NMR pulse - got it on the flip-flop branch from some of Serwans's pulse sequence upgrades

### DIFF
--- a/silq/pulses/pulse_sequences.py
+++ b/silq/pulses/pulse_sequences.py
@@ -525,7 +525,8 @@ class NMRPulseSequence(PulseSequenceGenerator):
             'stage_pulse': DCPulse('empty'),
             'NMR_pulse': SinePulse('NMR'),
             'NMR_pulses': ['NMR_pulse'],
-            'intermediate_pulses' : [],
+            'post_pulse': DCPulse('read', acquire=True),
+            'intermediate_pulses': [],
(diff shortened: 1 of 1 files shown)


silq/pulses/pulse_sequences.py
```diff
@@ -596,6 +603,10 @@ def add_NMR_pulses(self, pulse_sequence=None):
                     int_pulse, = self.add(intermediate_pulse)
                     t_connect(int_pulse['t_start'])
                     t_connect = partial(int_pulse['t_stop'].connect, offset=0)
+            elif self.NMR['post_pulse'] is not None:
```
maij, line 606: Do you only want the post-pulse to be added when there aren't any `NMR_pulses`?
I thought `post_pulse` would always be appended to the end.

RostyslavSavytskyy: No, (self.NMR['post_pulse'] is not None) - this line of code you are referring to is to insure that post_pulse is added at the end of all NMR pulses only if it exists.

maij: Ah I see, that makes sense because it's in the NMR settings dictionary, my mistake.

PR #212, opened 2020-04-24
Feature/multi sine pulse

RostyslavSavytskyy:
first implementation of MultiSinePulse in pulse types and interfaces. Will be tested soon and upgraded, don't be to harsh for now :)

Comment by Rosty:
this branch should be pulled after arbstudio_upgrade are already merged.


RostyslavSavytskyy:
Additional note: this branch should be pulled after E8267D_interface and arbstudio_upgrade are already merged.

silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -162,11 +166,6 @@ def get_additional_pulses(self, connections) -> List[Pulse]:
         if None in frequency_sidebands:
             frequency_sidebands.remove(None)
 
-        if frequency_sidebands or self.FM_mode() == 'IQ':
-            self.IQ_modulation._save_val('on')
-        else:
-            self.IQ_modulation._save_val('off')
-
```
maij, line 169: Why are these lines removed?

maij:
I started reviewing this, but I'm not really up-to-date with how the microwave source driver works (all the logic regarding IQ modulation is lost on me), so I think @nulinspiratie would be better suited for this one.

silq/instrument_interfaces/lecroy/ArbStudio1104_interface.py
```diff
@@ -420,20 +420,21 @@ def implement(self, sampling_rates, input_pulse_sequence, plot=False, **kwargs):
...
+            "Cannot implement multi sine pulse if the Arbstudio receives intermediary triggers"
 
         if isinstance(self.pulse.connection, SingleConnection):
             channels = [self.pulse.connection.output['channel'].name]
         else:
             raise Exception(f"No implementation for connection {self.pulse.connection}")
 
-        assert [frequency > 0 for frequency in self.pulse.frequencies], "All pulse frequencies must be larger than zero."
+        assert [frequency > 0 for frequency in self.pulse.frequencies], \
+            "All pulse frequencies must be larger than zero."
```
maij, line 437: If the ArbStudio is providing IQ modulation, then this prevents you from using the lower sideband and the carrier (frequency <= 0).
I don't think this is what you want.

silq/pulses/pulse_types.py
```diff
@@ -673,10 +673,11 @@ class MultiSinePulse(Pulse):
 
     Parameters:
         name: Pulse name
-        frequency: main Pulse frequency - !!! to be checked !!!
-        frequencies: list of pulse frequencies
-        phase: Pulse phase
+        frequency: main (local oscillator) Pulse frequency
```
maij, line 676: It's a nice addition being able to specify the carrier frequency in the pulse.
With a default value of `None`, do you expect that the carrier frequency is chosen by the interface that implements the pulse?

silq/instrument_interfaces/lecroy/ArbStudio1104_interface.py
RostyslavSavytskyy: well, generally we need just one sideband, which if IQ modulation is properly calibrated would be the only one giving power... unless I am missing something... Plus if we allow negative frequencies, I am not sure how it will work... sin(-|f|*t) - sounds confusing to me... I never worked with IQ modulation before, so don't have much experience... if something is off I am happy to listen if you have time to explain (WhatsApp/Teams or whatever is good for you)

silq/pulses/pulse_types.py
RostyslavSavytskyy: We always specify this frequency (haven't tested with None value, but would expect error), and then interface treats this as the main frequency of the pulse and determines carrier frequency depending on the pulse sequence, e.g. if it's the only pulse in pulse sequence, then carrier frequency = this main frequency; if we have another pulse at another frequency, then interface will determine carrier based on those two main frequencies

silq/pulses/pulse_types.py
```diff
@@ -668,6 +668,138 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
...
+
+        self.frequency = Parameter(initial_value=frequency, unit='Hz',
+                                   set_cmd=None, vals=vals.Numbers())
+        self.frequencies = Parameter(initial_value=frequencies, unit='Hz',
+                                     set_cmd=None, vals=vals.Lists())
+        self.phases = Parameter(initial_value=phases, unit='deg', set_cmd=None,
+                                vals=vals.Lists())
+        self.power = Parameter(initial_value=power, unit='dBm', set_cmd=None,
+                               vals=vals.Numbers())
+        self.amplitude = Parameter(initial_value=amplitude, unit='V',
```
nulinspiratie, line 721: Why does it also have a single amplitude?

silq/pulses/pulse_types.py
```diff
@@ -668,6 +668,138 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
...
+
+        assert len(self.amplitudes) == len(self.frequencies) == len(self.phases), f'Pulse {self.name} does not have '\
+                                                                                  f'equal number of amplitudes, ' \
+                                                                                  f'frequencies and phases.'
+        waveform = 0.0
+        for amp, freq, phase in zip(self.amplitudes, self.frequencies, self.phases):
+            waveform += amp * np.sin(2 * np.pi * (freq * t + phase / 360))
+        waveform += self.offset
+        # We need to insure the waveform is limited by +- 1V:
+        waveform = waveform/len(self.frequencies)
```
nulinspiratie, line 799: I don't agree with this line. The pulse itself shouldn't by default be limited to +-1V, this is only the case if it is used for IQ modulation. If I create the sum of two Sine pulses, each with an amplitude of 1V, I don't expect each pulse to suddenly have an amplitude of 0.5V

silq/pulses/pulse_types.py
```diff
@@ -668,6 +668,138 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
...
+                 power: float = None,
+                 offset: float = None,
+                 frequency_sideband: float = None,
+                 sideband_mode: float = None,
+                 phase_reference: str = None,
+                 **kwargs):
+
+        super().__init__(name=name, **kwargs)
+
+        self.frequency = Parameter(initial_value=frequency, unit='Hz',
```
nulinspiratie, line 713: Is this for the carrier frequency? If so, it needs to have a different name, else it will be too confusing. I also don't know if it's a good idea to specify the carrier frequency in a pulse, as it may lead to problems later on. For instance, what if different pulses request different carrier frequencies? I feels more like a property of the microwave interface than of the pulse, but it would be good to discuss this

silq/instrument_interfaces/lecroy/ArbStudio1104_interface.py
maij: I'll try to answer this here, but if it's still unclear then we can have a chat.

Basically I'm saying that you can use IQ modulation to target either the upper or lower sideband, offset from the carrier frequency by the frequency of the IQ terms, depending on whether the IQ terms are rotating clockwise or counter-clockwise. 

So the microwave interface tries to implement a `MultiSinePulse` it will create additional pulses for the ArbStudio, which are also `MultiSinePulse`s. In the `MultiSinePulseImplementation` of the ArbStudio, you assert that the frequencies of each component Sine must be greater than 0, but this prevents you from utilising the clockwise-type of IQ modulation, i.e. it prevents you from modulating the microwave output frequency below that of the carrier frequency. You've also prevented having a constant term (i.e. frequency = 0) which would output at the carrier frequency.

I understand that you generally only use the upper sideband, but I see no reason to impose this limitation in the driver. If you only want to use the upper sideband, then you can just set your carrier frequency accordingly.

silq/pulses/pulse_types.py
maij: In that case, since `frequency` is a required argument you shouldn't give it the default value of `None`. I also agree with Serwan that it should be renamed to something more informative.

silq/instrument_interfaces/lecroy/ArbStudio1104_interface.py
RostyslavSavytskyy: Just to confirm I understand correctly - if the frequency w < 0, then our the main modulated signal will be just on the lower sideband instead of the upper sideband, i.e. we just switch from sin(w_LO+w) to sin(w_LO-|w|)? If so, then I agree, ok, thanks, will correct this.

silq/pulses/pulse_types.py
RostyslavSavytskyy: What kind of default value then should I set? I copied this from other pulses - they all have None default value. I don't see the reason setting to something by default (magnetic field always varies), unless you mean I should add some kind of condition/demand on this frequency...
Concerning the renaming  - I didn't rename on purpose, since when microwave interface is checking pulses in pulse_sequence it works with "frequency" parameter of the pulse - I just wanted to be compatible with the interface. But, I suppose, of course, I could change interface appropriately for this pulse...

silq/pulses/pulse_types.py
RostyslavSavytskyy: I left it to be able to define microwave power from this amplitude, but I don't think it is implemented in the interface plus we never use this, so I agree, I will remove this

silq/pulses/pulse_types.py
RostyslavSavytskyy: True, I will move these conditions into interface

silq/pulses/pulse_types.py
RostyslavSavytskyy: This frequency is not exactly the carrier frequency. If we have only one pulse in pulse sequence, and frequency carrier choice is set to 'center', then yes - it acts as a carrier frequency. However, if we have more pulses, e.g. SinePulse for ESR1 and MultiSinePulse for ESR2, then microwave interface treats this frequency as the main frequency of the MultiSinePulse similarly to FrequencyRampPulse. In this case, depending on the carrier choice, the carrier frequency will be determined from those two ESR1 and ESR2 frequencies of the pulses. So, in short, I didn't rename on purpose, since when microwave interface is checking pulses in pulse_sequence it works with "frequency" parameter of the pulse - I just wanted to be compatible with the interface.

silq/instrument_interfaces/lecroy/ArbStudio1104_interface.py
RostyslavSavytskyy: Corrected.

silq/pulses/pulse_types.py
RostyslavSavytskyy: Corrected.

silq/pulses/pulse_types.py
RostyslavSavytskyy: corrected.

silq/pulses/pulse_types.py
RostyslavSavytskyy: Let's discuss this, I am not sure what kind of issues you have in mind

RostyslavSavytskyy: I looked again at all of this and generally I agree - we can remove "frequency" from the MultiSine Pulse with a few adjustments to the interface min/max frequency determination, let's shortly discuss this and I will remove it.

silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -206,7 +240,8 @@ def get_additional_pulses(self, connections) -> List[Pulse]:
             "Maximum FM frequency deviation is 80 MHz if FM_mode == 'ramp'. " \
             f"Current frequency deviation: {self.frequency_deviation()/1e6} MHz"
 
-        if frequency_sidebands or (self.FM_mode() == 'IQ' and min_frequency != max_frequency):
+        if frequency_sidebands or (self.FM_mode() == 'IQ' and
+                                   ((self.frequency_deviation != 0) or (multiple_frequencies is not None))):
```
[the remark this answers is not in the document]
RostyslavSavytskyy: corrected

silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -164,8 +194,12 @@ def get_additional_pulses(self, connections) -> List[Pulse]:
...
         min_frequency = max_frequency = None
+        multiple_frequencies = None
         for pulse in self.pulse_sequence:
             frequency_deviation = getattr(pulse, 'frequency_deviation', None)
+            multi_freq = getattr(pulse, 'frequencies', None)
+            if multi_freq is not None:
+                multiple_frequencies = multi_freq
             frequency_sideband = pulse.frequency_sideband
 
             pulse_min_frequency = pulse_max_frequency = pulse.frequency
```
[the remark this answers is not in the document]
RostyslavSavytskyy: corrected

silq/pulses/pulse_types.py
RostyslavSavytskyy: corrected

silq/pulses/pulse_types.py
RostyslavSavytskyy: corrected

RostyslavSavytskyy:
Removed frequency parameter and updated microwave interface, should be good to go

silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -381,17 +420,21 @@ def get_additional_pulses(self, interface: InstrumentInterface):
                           t_start=self.pulse.t_start - interface.envelope_padding(),
                           t_stop=self.pulse.t_stop + interface.envelope_padding(),
                           frequency=frequency_IQ,
-                          amplitude=1,
-                          phase=0,
+                          amplitude=1 + interface.I_amplitude_correction(),
+                          phase=0 + interface.I_phase_correction(),
```
maij, line 424: Should get rid of the 0 now.
```suggestion
                          phase=interface.I_phase_correction(),
```

silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -441,9 +488,75 @@ def get_additional_pulses(self, interface: InstrumentInterface):
...
+                        amplitude=interface.marker_amplitude(),
+                        connection_requirements={
+                            'input_instrument': interface.instrument_name(),
+                            'input_channel': 'trig_in'})]
+
+        assert (interface.IQ_modulation() == 'on') and \
+               (interface.FM_mode() == 'IQ'), 'FM_mode should be IQ and IQ_modulation should be ON ' \
+                                              'for MultiSinePulse.'
+
+        # To insure the waveform is limited by +- 1V:
```
maij, line 519:
```suggestion
        # To ensure the waveform is limited to +- 1V:
```

silq/instrument_interfaces/keysight/E8267D_interface.py
RostyslavSavytskyy: corrected

silq/instrument_interfaces/keysight/E8267D_interface.py
RostyslavSavytskyy: corrected

RostyslavSavytskyy:
all good now

nulinspiratie:
@RostyslavSavytskyy The merge conflict needs to be resolved. After that, I'll have a good look through it. I think this PR can be merged soon

RostyslavSavytskyy:
The conflict should be resolved now. Have a look.

silq/pulses/pulse_types.py
```diff
@@ -667,6 +667,121 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
...
 
 
+class MultiSinePulse(Pulse):
+    """MultiSinusoidal pulse
+
+    Parameters:
+        name: Pulse name
+        frequencies: list of Pulse frequencies
+        phases: list of Pulse phases
+        amplitudes: list of Pulse amplitudes
```
nulinspiratie, line 677: From what I understand, the meaning of amplitudes depends on the interface.
For the arbstudio, these amplitudes correspond to the actual amplitudes of the sine pulses.
For the microwave source, these amplitudes are scaled such that the total amplitude is 1.

This should be clarified in the documentation. Don't refer to specific interfaces, but add that if the interface uses the power, amplitudes are used for IQ modulation and scaled such that the sum is 1. Else the amplitudes are not scaled.

silq/pulses/pulse_types.py
```diff
@@ -667,6 +667,121 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
...
+        power: Pulse power
+        offset: amplitude offset (0 by default)
+        frequency_sideband: Mixer sideband frequency (off by default)
+        sideband_mode: Sideband frequency to apply. This feature must
```
nulinspiratie, line 681: frequency_sideband and sideband_mode have nearly identical documentation. What do they exactly correspond to, and how do they differ?

silq/pulses/pulse_types.py
```diff
@@ -667,6 +667,121 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
...
+        assert self.amplitudes is not None, f'Pulse {self.name} does not have ' \
+                                            f'specified amplitudes.'
+        assert self.frequencies is not None, f'Pulse {self.name} does not have ' \
+                                             f'specified frequencies.'
+        assert self.phases is not None, f'Pulse {self.name} does not have specified phases.'
+
+        assert len(self.amplitudes) == len(self.frequencies) == len(self.phases), \
+            f'Pulse {self.name} does not have equal number of amplitudes, frequencies and phases.'
+
+        waveform = 0.0
```
nulinspiratie, line 777: create an empty array. Otherwise if amplitudes etc. is an empty list, it would return a float instead of an array

nulinspiratie (changes requested):
Small changes, mostly documentation. Then it's good to go

silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -73,6 +77,32 @@ def __init__(self, instrument_name, **kwargs):
...
+                           vals=vals.Numbers(min_value=-1, max_value=0),
+                           docstring="To calibrate the amplitude of I component independent of Q component."
+                                     "Takes values from -1 to 0V. Only for FM_mode = 'IQ'.")
+        self.add_parameter('Q_amplitude_correction',
+                           unit='V',
+                           set_cmd=None,
+                           initial_value=0,
+                           vals=vals.Numbers(min_value=-1, max_value=0),
+                           docstring="To calibrate the amplitude of Q component independent of I component."
+                                     "Takes values from -1 to 0V. Only for FM_mode = 'IQ'.")
```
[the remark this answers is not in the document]
RostyslavSavytskyy: Added more description, have a look. -1 to 0 restriction is to make sure we don't apply above 1V to I/Q inputs.
The phase correction aims to correct any possible phase offset deviations between I and Q pulses, e.g. instead of 90 degrees we might have (instrumentally) 89 degrees phase offset, which should be corrected by these parameters and it should not depend on the frequency a lot, at least for those frequencies that you calibrate this for...

silq/pulses/pulse_types.py
```diff
@@ -667,6 +667,121 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
         return waveform
 
 
+class MultiSinePulse(Pulse):
+    """MultiSinusoidal pulse
```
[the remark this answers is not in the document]
RostyslavSavytskyy: done

silq/pulses/pulse_types.py
```diff
@@ -667,6 +667,121 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
...
+
+    Parameters:
+        name: Pulse name
+        frequencies: list of Pulse frequencies
+        phases: list of Pulse phases
+        amplitudes: list of Pulse amplitudes
+        power: Pulse power
```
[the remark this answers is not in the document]
RostyslavSavytskyy: done

silq/pulses/pulse_types.py
RostyslavSavytskyy: done

silq/pulses/pulse_types.py
RostyslavSavytskyy: done

silq/pulses/pulse_types.py
RostyslavSavytskyy: Honestly, I have never used those and don't really understand how they work. So I just copied from other pulses since it is required for the microwave interface. Can discuss in the office...

RostyslavSavytskyy:
Have a look now

silq/pulses/pulse_types.py
nulinspiratie: yep I see it's the same issue in other pulses. We'll discuss later

nulinspiratie (approved):
good to go

RostyslavSavytskyy:
Thanks!

PR #215, opened 2020-04-29
fixed factor 2 missing in FrequencyRampPulse

RostyslavSavytskyy:


nulinspiratie:
@maij @RostyslavSavytskyy Did anyone actually check that the frequency was indeed off by a factor 2?

nulinspiratie:
As in look at the pulse sequence and verify that the final frequency is lower than it should be?

nulinspiratie:
Nevermind, I verified and it was indeed off

RostyslavSavytskyy:
I can just add that with this factor of 2, the adiabatic peak had the correct width

PR #220, opened 2020-05-13
Fix: Check if using ipython before registering magic

nulinspiratie:
Originally the only check was if `ipykernel` was in `sys.modules`.
While this usually works in detecting if we're using ipython, it fails when a new process is started using multiprocessing. In this situation `ipykernel` exists, but the process is not using ipython.

This modification was needed for https://github.com/nulinspiratie/Qcodes/pull/79


nulinspiratie:
I believe so, yes

PR #221, opened 2020-05-25
Simgui will not be placed off-screen for small resolution displays

maij:
Quick fix, ready to go


nulinspiratie:
Get a bigger monitor though

PR #226, opened 2020-06-16
feat: New blips analyses using maximum likelihood estimation

nulinspiratie:
Beta testing at the moment

### DIFF
--- a/silq/analysis/tunneling_times.py
+++ b/silq/analysis/tunneling_times.py
@@ -638,15 +638,35 @@ class TunnelTimesAnalysis:
 
 def analyse_tunnel_times_measurement(
     data=None,
+    traces=None,
     tunnel_times_analyses=None,
     threshold_voltage=150e-3,
     t_skip=0,
     sample_rate=200e3,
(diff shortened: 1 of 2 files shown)


silq/analysis/tunneling_times.py
```diff
@@ -193,5 +210,677 @@ def t_read_optimal(tau_up, tau_down):
     """Optimal read duration"""
     return 1/(1/tau_down - 1/tau_up) * np.log(tau_up / tau_down)
 
+
 def contrast_optimal(t, tau_up, tau_down):
-    return np.exp(-t/tau_down) - np.exp(-t/tau_up)
\ No newline at end of file
+    return np.exp(-t/tau_down) - np.exp(-t/tau_up)
```
RostyslavSavytskyy, line 215: shouldn't it be with plus sign instead of minus?

RostyslavSavytskyy (approved):
I think that this pull request is missing some docstrings since most of the procedures are new and unclear, but don't see any bugs for now, so can approve

PR #227, opened 2020-06-16
added phase to FrequencyRampPulse

RostyslavSavytskyy:


silq/pulses/pulse_types.py
```diff
@@ -603,7 +602,7 @@ def __init__(self,
                                          set_cmd=None, vals=vals.Enum('relative',
                                                                       'absolute'))
         self._connect_parameters_to_config(
-            ['frequency', 'phase', 'power', 'amplitude', 'phase', 'offset',
```
nulinspiratie, line 606: Why is phase removed?

silq/pulses/pulse_types.py
```diff
@@ -684,6 +683,11 @@ class FrequencyRampPulse(Pulse):
             initial frequency or stay at current frequency. Useful if the
             pulse doesn't immediately stop at the end (this depends on how
             the corresponding instrument/interface is programmed).
+        phase: Pulse phase. By default is set to zero.
+        phase_reference: What point in the the phase is with respect to.
+            Can be two modes:
+                - 'absolute': phase is with respect to `Pulse.t_start`.
```
nulinspiratie, line 689: Absolute and relative are swapped here

silq/pulses/pulse_types.py
RostyslavSavytskyy: because it's repeated twice there

silq/pulses/pulse_types.py
RostyslavSavytskyy: I copied this from SinePulse - do you have different definitions for different pulses?

RostyslavSavytskyy: Swapped them and also for SinPulse as well

PR #228, opened 2020-06-23
Feature/easy access to fit parameters and function call

nulinspiratie:
This PR enables easy access to fitted parameters via `fit['{parameter_name}']`.
It also allows easy evaluation of the fitted function using `fit(sweep_values)`, which by default uses the fitted parameters. They can be overridden by passing kwargs

### DIFF
--- a/silq/analysis/fit_toolbox.py
+++ b/silq/analysis/fit_toolbox.py
@@ -22,7 +22,10 @@ class Fit():
     after which it will fit the data to `Fit.fit_function`.
 
     Note:
-        The fitting routine uses lmfit, a wrapper package around scipy.optimize.
+        - The fitting routine uses lmfit, a wrapper package around scipy.optimize.
+        - Fitted parameters can be accessed via ``fit['{parameter_name}']``
+        - The fit function can be evaluated with the fitted parameter values
+          using ``fit({sweep_values})``
     """
     plot_kwargs = {'linestyle': '--', 'color': 'cyan', 'lw': 3}
     sweep_parameter = None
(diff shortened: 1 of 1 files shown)


silq/analysis/fit_toolbox.py
```diff
@@ -43,6 +46,38 @@ def __init__(self, fit=True, print=False, plot=None, **kwargs):
             if fit:
                 self.perform_fit(print=print, plot=plot, **kwargs)
 
+    def __getitem__(self, item) -> float:
+        """Retrieve fitted parameter value"""
+        return self.fit_result.best_values[item]
```
maij, line 51: If I do something like this:
```python
fit = SineFit()
fit['frequency']
```
I get the following error
```python-traceback
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-18-827926a6d6a1> in <module>
      1 fit = SineFit()
      2 
----> 3 fit['frequency']

~/Documents/github/SilQ/silq/analysis/fit_toolbox.py in __getitem__(self, item)
     49     def __getitem__(self, item) -> float:
     50         """Retrieve fitted parameter value"""
---> 51         return self.fit_result.best_values[item]
     52 
     53     def __call__(

AttributeError: 'NoneType' object has no attribute 'best_values'
```

This case is quite trivial to figure out why it failed, but I think this should be captured and a more informative error given.

e.g. 
```suggestion
    def __getitem__(self, item) -> float:
        """Retrieve fitted parameter value"""
        if self.fit is not None:
             return self.fit_result.best_values[item]
        else:
              raise RuntimeError("No fit result to get parameters from.") # Probably not a key or index error, but something more informative than this would be best
```

nulinspiratie: Done (without the comment)

PR #232, opened 2020-07-27
Feature/iqdc phase control

RostyslavSavytskyy:
IQ DC phase control for the Sine pulse to avoid jitters/errors during triggering of each pulse

### DIFF
--- a/silq/instrument_interfaces/keysight/E8267D_interface.py
+++ b/silq/instrument_interfaces/keysight/E8267D_interface.py
@@ -154,6 +154,11 @@ class E8267DInterface(InstrumentInterface):
                                      'cannot be directly set, but is determined '
                                      'by FM_mode and whether pulses have '
                                      'frequency_sideband not None')
+        # TODO expand docstring
+        self.add_parameter('force_IQ',
+                           initial_value=False,
+                           vals=vals.Bool(),
+                           docstring='Whether to enforce IQ modulation.')
         self.add_parameter('FM_mode',
                            set_cmd=None,
                            initial_value='ramp',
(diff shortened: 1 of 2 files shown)


RostyslavSavytskyy:
Since multi_sine branch is merged, we can proceed with this pull request. Updated docstring, should be good to go.

maij (commented):
I don't really understand why this should be necessary, originally when `frequency_IQ == 0`, the interface would request a I/Q SinePulses with frequency 0 and the relevant phases. This is exactly the same as the DC pulses you've implemented. You can verify this by constructing the two pulses (the DC pulse and the SinePulse) and calling get_voltage for their durations. 

```python
from silq.pulses import SinePulse, DCPulse

phase = 32.7654
t_start = 0
t_stop = 100e-6
t = np.linspace(t_start, t_stop, 10000)

dc_pulse = DCPulse('sideband_quadrature', amplitude=np.sin(2 * np.pi * phase / 360), t_start=t_start, t_stop=t_stop)
sine_pulse = SinePulse('sideband_quadrature', amplitude=1, phase=phase, frequency=0, t_start=t_start, t_stop=t_stop)

assert np.all(dc_pulse.get_voltage(t) == sine_pulse.get_voltage(t))
```

---
Also, the DCpulses don't have the IQ amplitude/phase corrections, should these be added if this is desirable?

RostyslavSavytskyy:
I see your point, yes, you are correct, I think we just wanted to make sure that we get DC pulses there. I will have a look/test this quickly and probably remove this explicit option.

RostyslavSavytskyy:
Looked at this again: we could use SinePulse with frequency 0 to implement DCPulses for I and Q components, however due to current Arbstudio/AWG implementation, the SinePulse is required to have >100 Hz frequency. This limitation comes from the maximum number of waveform points per period, which is possible to load onto the AWG (small frequency -> long period -> lots of waveform points). On the other hand, if frequency is small, we don't need the whole period e.g. of 1s, but just a small part of it, e.g. 1us, which means we need to implement the case where Arbstudio uses "stepped" mode for SinePulses of small frequency. Once this is implemented, then IQ DCPulse implementation can be merged with the general SinePulse implementation.

nulinspiratie:
I'm undecided about whether the SinePulse or DCPulse is a better idea, I can see the advantages of both.
I like the PR in general, force_IQ is a good addition. One thing I would suggest to change is the docstring of `force_IQ`, which is a bit cryptic. If help is needed let me know

RostyslavSavytskyy:
Modified the docstring, have a look now

nulinspiratie:
Better, but IMO the reasoning why this would be useful is unnecessary, also because there are many different reasons why this may be needed. What about:

`Force IQ modulation to be enabled, even when outputting only a single frequency`

RostyslavSavytskyy:
All good now, also added phase correction like Mark suggested

nulinspiratie (approved):
Good job Rosty, I'm proud of you

PR #233, opened 2020-07-28
Add single waveform pulse and implementations to some drivers

RostyslavSavytskyy:
Single waveform pulse contains all pulses of the same frequency but different phases and durations - designed to have a single waveform for one trigger - should reduce trigger jitters/errors, etc.

Note: should be merged after IQDC_phase_control pull requests


maij (commented):
Just a comment before I review this, what is the 'exponential saturation' ramp pulse for? Are you trying to implement an exponential chirp?

RostyslavSavytskyy:
Yes, correct. At the moment this SingleWaveformPulse is fairly fresh and we tried to implement Stark shift correction as well. Since we saw linear and exponential-saturation dependencies for it, we implemented both options, but the option for exponential-saturation is not sustainable as it is right now. Will improve it if we use it indeed or remove at all. We actually started discussing with Tim about rewriting this pulse (or adding another one), where we feed it with the function formula, similar to the AWG Pulse (or maybe we just modify AWG pulse a bit). But in this case it will also have multiple such pulses in one single waveform of varying phase, duration, frequency, etc.

RostyslavSavytskyy:
I think yes, Irene already started using this, so we might integrate this into SilQ completely.

silq/instrument_interfaces/keysight/Keysight_81180A_interface.py
```diff
@@ -512,7 +515,7 @@ def add_pulse(self, pulse, t, channel, sample_rate):
             )
```
maij, line 515: These changes are from a different PR. Need to remove before merging.

silq/pulses/pulse_types.py
```diff
@@ -848,6 +847,249 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
...
+                        waveform[idx_list] += amplitude * np.sin(
+                            2 * np.pi * (frequency * t[idx_list] +
+                                         phase / 360))
+                    elif self.AM_type == 'gauss':
+                        t_start = sum(self.durations[:idx])
+                        t_end = sum(self.durations[:idx + 1])
+                        peak_t = (t_start + t_end) / 2
+                        waveform[idx_list] += amplitude * np.exp(
+                            -8 * (t[idx_list] - peak_t) ** 2 / (duration ** 2)) * np.sin(
+                            2 * np.pi * (frequency * t[idx_list] + phase / 360))
```
maij, line 1041: There is no control for the sharpness of the Gaussian, you just have a fixed -8 factor in here.
Is this standard for pulse-shaping?

silq/pulses/pulse_types.py
```diff
@@ -848,6 +847,249 @@ def get_voltage(self, t: Union[float, Sequence]) -> Union[float, np.ndarray]:
...
+                    if self.pulse_type == 'sine':
+                        if self.AM_type == 'square':
+                            waveform += amplitude * np.sin(2 * np.pi * (frequency * t + phase / 360))
+                        elif self.AM_type == 'gauss':
+                            t_start = sum(self.durations[:idx])
+                            t_end = sum(self.durations[:idx + 1])
+                            peak_t = (t_start + t_end) / 2
+                            waveform += amplitude * np.exp(
+                                -8 * (t - peak_t) ** 2 / (duration ** 2)) * np.sin(
+                                2 * np.pi * (frequency * t + phase / 360))
```
maij, line 1074: See above.

maij (commented):
Looks good, just one small question before approving.

silq/pulses/pulse_types.py
RostyslavSavytskyy: To be honest, I have no idea where -8 factor came from, maybe some typo or something

maij (approved):

PR #234, opened 2020-07-29
Feature/int fm modulation

RostyslavSavytskyy:
Added an option for microwave source's internal modulation of adiabatic ramp pulses


RostyslavSavytskyy:
So I disentangled this branch from SingleWaveform, added comments, and should be ready for a review and merge. Have a look

nulinspiratie (approved):
Looks good, go ahead

PR #235, opened 2020-07-30
Feature: add option to specify  acquisition channel label in acquisition parameters

maij:
I'm trying to make sure that every acquisition parameter has the option to specify which acquisition channel it will analyse for information.

This became apparent when measuring 2 devices simultaneously, where the DCParameter could only analyse the channel 'output'.

### DIFF
--- a/silq/parameters/acquisition_parameters.py
+++ b/silq/parameters/acquisition_parameters.py
@@ -67,7 +67,6 @@ class AcquisitionParameter(SettingsClass, MultiParameter):
         properties_attrs: attributes to match with
             ``silq.config.properties`` (see notes below).
         save_traces: Save acquired traces to disk
-        channel_label: The layout acquisition channel label used for analysis.
         **kwargs: Additional kwargs passed to ``MultiParameter``
 
     Parameters:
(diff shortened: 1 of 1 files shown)


nulinspiratie:
I like it. Only comment is that perhaps instead of passing the channel label as a kwarg, we can instead have it as a class attribute. This way it can be changed for the entire class. Also, it can later be overwritten for specific object instances

maij:
> I like it. Only comment is that perhaps instead of passing the channel label as a kwarg, we can instead have it as a class attribute. This way it can be changed for the entire class. Also, it can later be overwritten for specific object instances
Hmm, it sounds like this is something that should be incorporated into the experiment config at some point down the road. One potential pitfall is a situation like the following, where you might want to subclass another AcquisitionParameter:

```python
class A:
    var = 'test'
    
class B(A):
    pass

a = A()
b = B()

A.var = 'hi'
print(b.var)
```
Output: 'hi'

Currently the only parameters that do this are the `EDSRParameter` and the `NeuralRetuneParameter`, but I believe these might be changed in the near future.

PR #243, opened 2020-08-28
Speed up segmenting of traces in ATS

nulinspiratie:
I noticed that segmenting traces in the ATS (when `pulse.average_mode == 'segment:{N}') was slow, especially when using the DCSweepPlot.
Here we speed this process up quite a bit


maij (approved):
Looks good.

PR #244, opened 2020-08-28
Add timing of DCSweepPlot and minor code speedup

nulinspiratie:
Small improvements to DCSweepPlot:
- Add `DCSweepPlot.timings` which gives information on how long acquisition and plotting takes
- Call get_latest from voltage parameters to speed up refreshing


maij (approved):
Seems fine.

PR #245, opened 2020-09-03
add envelope_IQ to keysight e8267D interface

nulinspiratie:
Adds the parameter `microwave_interface.envelope_IQ` to the Keyisght E8267D microwave interface.
By default (`envelop_IQ=True`), the envelope padding is applied to the IQ inputs, while the marker pulses are the length of the actual pulse.
When `envelope_IQ = False`, the envelope padding is instead applied to the marker pulses, and the IQ inputs are only nonzero for the duration of the pulses.
This feature is useful if you want multiple pulses that are more closely spaced than the envelope padding.

Note that to avoid pulse overlaps, any overlapping marker pulses are merged into a single longer marker pulse.

@RostyslavSavytskyy I think we should merge this PR before #233 since it is a fairly small PR and could otherwise lead to conflicts

### DIFF
--- a/silq/instrument_interfaces/keysight/E8267D_interface.py
+++ b/silq/instrument_interfaces/keysight/E8267D_interface.py
@@ -88,7 +88,7 @@ class E8267DInterface(InstrumentInterface):
                       "I and Q pulses are extended by the envelope padding "
                       "before (after) the pulse start (stop) time. "
                       "If False, the marker pulse starts before (after) the pulse start (stop) "
-                      "time, and the IQ pulses start (stop) at the pulse start (stop) time."
+                      "time, and the IQ pulses starts (stops) at the pulse start (stop) time."
                       "This means that there might be some leakage at the carrier frequency "
                       "when the IQ is zero, but the marker pulse is still high."
         )
@@ -203,6 +203,8 @@ class E8267DInterface(InstrumentInterface):
                                if pulse.frequency_sideband is not None else None
                                for pulse in self.pulse_sequence}
 
+        # Ramp mode does not work with envelope IQ since the output can only be
+        # turned off via gating (0V on ramp input corresponds to a pulse at carrier)
         assert self.FM_mode() != 'ramp' or self.envelope_IQ(), \
             "Cannot use 'ramp' FM mode while envelope_IQ == False"
 
--- a/silq/instrument_interfaces/keysight/E8267D_interface.py
+++ b/silq/instrument_interfaces/keysight/E8267D_interface.py
@@ -6,6 +6,8 @@ from silq.pulses import Pulse, DCPulse, DCRampPulse, SinePulse, \
     FrequencyRampPulse, MarkerPulse, PulseImplementation
 
 from qcodes.utils import validators as vals
+from qcodes.instrument.parameter import Parameter
+
 
 class E8267DInterface(InstrumentInterface):
     """ Interface for the Keysight E8267D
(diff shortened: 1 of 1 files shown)


silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -77,6 +79,19 @@ def __init__(self, instrument_name, **kwargs):
...
                                      "FM_mode = 'IQ'.")
+        self.envelope_IQ = Parameter(
+            set_cmd=None,
+            vals=vals.Bool(),
+            initial_value=True,
+            docstring="Apply the envelope padding to the IQ pulse instead of marker pulse."
+                      "If True, the marker pulse length equals the pulse length, and the "
+                      "I and Q pulses are extended by the envelope padding "
+                      "before (after) the pulse start (stop) time. "
+                      "If False, the marker pulse starts before (after) the pulse start (stop) "
```
RostyslavSavytskyy, line 90: starts (stops)

silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -188,6 +203,9 @@ def get_additional_pulses(self, connections) -> List[Pulse]:
                                if pulse.frequency_sideband is not None else None
                                for pulse in self.pulse_sequence}
 
+        assert self.FM_mode() != 'ramp' or self.envelope_IQ(), \
+            "Cannot use 'ramp' FM mode while envelope_IQ == False"
+
```
RostyslavSavytskyy, line 208: what ramp mode has to do with envelope_padding for IQ modulation?

RostyslavSavytskyy (changes requested):
The idea is a bit unclear, (can discuss in the office), but at the moment it is implemented only for Sine Pulse, should it be similar for other pulses?

silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -258,6 +276,23 @@ def get_additional_pulses(self, connections) -> List[Pulse]:
...
+                if pulse.t_start <= current_pulse.t_stop:
+                    # Marker pulse overlaps with previous pulse
+                    current_pulse.t_stop = max(pulse.t_stop, current_pulse.t_stop)
+                else:
+                    # Pulse starts after previous pulse
+                    merged_marker_pulses.append(pulse)
+                    current_pulse = pulse
+            nonmarker_pulses = [p for p in additional_pulses if not isinstance(p, MarkerPulse)]
+            additional_pulses = [*merged_marker_pulses, *nonmarker_pulses]
+
```
RostyslavSavytskyy, line 295: I am a bit afraid that if envelope_padding has "large" value, and general (not marker) pulses have "small" inter-delays, there will be lots of overlaps there since you sort only marker pulses...

RostyslavSavytskyy (changes requested):

silq/instrument_interfaces/keysight/E8267D_interface.py
nulinspiratie: fixed

silq/instrument_interfaces/keysight/E8267D_interface.py
nulinspiratie: added note:
        # Ramp mode does not work with envelope IQ since the output can only be
        # turned off via gating (0V on ramp input corresponds to a pulse at carrier)

silq/instrument_interfaces/keysight/E8267D_interface.py
nulinspiratie: I don't follow your reasoning, we should discuss

RostyslavSavytskyy:
Ok, I checked your changes again and: 
1) I still don't completely understand the advantage of adding this feature (let's discuss whenever you have time)
2) Shouldn't we add the same envelope_padding control for other pulse implementations and not only for SinePulse?

silq/instrument_interfaces/keysight/E8267D_interface.py
nulinspiratie: Is it clear now?

silq/instrument_interfaces/keysight/E8267D_interface.py
nulinspiratie: Can we resolve this conversation?

nulinspiratie:
@RostyslavSavytskyy what do you think now?

silq/instrument_interfaces/keysight/E8267D_interface.py
RostyslavSavytskyy: yes

silq/instrument_interfaces/keysight/E8267D_interface.py
RostyslavSavytskyy: yes

RostyslavSavytskyy (approved):

PR #246, opened 2020-09-19
Ignore .ipynb_checkpoints when running scripts in folder

nulinspiratie:


maij (approved):
I thought I fixed this one a while back, good catch.

PR #247, opened 2020-09-21
added deletion of Q character for RB sequences

RostyslavSavytskyy:
In RB sequences for Randomized Benchmarking there is an additional Q character that is needed to properly use pygsti analysis of the results later. But to measure first, we need to get rid of it.

### DIFF
--- a/silq/tools/circuit_tools.py
+++ b/silq/tools/circuit_tools.py
@@ -22,6 +22,7 @@ def convert_circuit(circuit, target_type: Type = str):
     elif isinstance(circuit, str):
         # Replace [] by identity gate
         expanded_circuit = circuit.replace('[]', 'Gi')
+        expanded_circuit = circuit.replace('Q', '')
 
         # Expand any expressions ({gates})^exponent
         expand_subcircuit = lambda match: match.group(0).replace(


nulinspiratie:
What is the meaning of the Q character? is it a gate, or what else is it?

RostyslavSavytskyy:
During generation of random circuits you need to add "qubit labels" (generally it works with character "Q" as a label for one qubit, for two have not tried, might need number as well, i.e. Q1, Q2, etc.). If you skip this label, the pygsti analysis of the data will not work - the final file with the measured results should be identical to what they create in the beginning with zeros, except you change zeros to measured values.

nulinspiratie:
@RostyslavSavytskyy please don't merge branches without approvals. 
There is probably a better way to implement this change, and it is also missing a comment clarifying it. I will revert for now and we can discuss later

RostyslavSavytskyy:
It showed to me as if it was approved - so I merged, otherwise of course I wouldn't do this.

nulinspiratie:
Yeah all good. I think the difference is that because you're merging into a branch that is not the master branch, you can merge without an approval. So it may have looked like an approval

Issue #104, opened 2018-07-03
CombinedConnection cannot be used with multiple output instruments.


maij:
Issues reside in pulse targeting within the layout.

PR #250, opened 2020-09-24
Interfaces/pcdds auto advance

nulinspiratie:
Add feature for pulses to auto-advance after fixed interval. Requires newest PCDDS image

Linked to https://github.com/nulinspiratie/Qcodes/pull/104

### DIFF
--- a/silq/instrument_interfaces/keysight/fpga_interfaces/PCDDS_interface.py
+++ b/silq/instrument_interfaces/keysight/fpga_interfaces/PCDDS_interface.py
@@ -63,14 +63,37 @@ class PCDDSInterface(InstrumentInterface):
         """
         assert self.use_trig_in(), "Interface not yet programmed for pxi triggering"
         # Get list of unique pulse start and stop times
-        t_list = self.pulse_sequence.t_list
-        trigger_pulses = [TriggerPulse(t_start=t,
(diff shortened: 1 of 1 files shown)


maij (commented):
I'm a bit confused about this PR, `auto_advance` is a Parameter of the interface, but it's never used. The implementation relies on individual pulses having the attribute `auto_advance` which is non-standard.

nulinspiratie:
@maij Perhaps the label "ready for review" should be removed, as it was added in a rush.
The problem here is that you may not want the entire pulse sequence to use auto_advance, but only a portion of it. In fact, I think that's the more common scenario. For this reason having a global interface parameter "auto_advance" probably isn't the right move. I think that parameter should be deleted.
On the other end of the spectrum is optionally setting an attribute "auto_advance" per pulse, though this of course feels slightly hacky.
I think the best solution in the end is to use the attribute "flags" of pulse sequences, which may have a key "auto_advance" This way nested pulse sequences can inherit properties from the parent pulse sequence. I believe I already implemented something like this in the 81180A AWG. We could also consider adding the attribute "flags" to the Pulse, which could also have the key "auto_advance". This way we could also set auto_advance on a pulse-by-pulse basis.

nulinspiratie:
Closed because I want to start over

PR #255, opened 2020-10-27
Add DummyPulse

nulinspiratie:
This pulse will be ignored by the layout. Useful if you for instance need a delay between pulses


RostyslavSavytskyy (approved):
Looks nice, but how would you use it? Like instead of setting inter-delay to some value, you add DummyPulse to some duration and set inter-delay to zero?

nulinspiratie:
@RostyslavSavytskyy yeah that's an option. you can also combine it with an inter_delay if you like, it's really just for when you want to pretend there's a pulse somewhere

PR #257, opened 2020-11-03
Fix: AWG indexing issues and inverted connection

nulinspiratie:
Fixes another minor issue regarding keysight indexing caused by `default_sampling_rates` still having hard-coded zero-based indexing.

Also adds the option for triggering lines that have an inverted connection (e.g. Pulseblaster DDS)


RostyslavSavytskyy (approved):
looks good to me

PR #259, opened 2020-11-04
2sideband modulation option and settings dict for MultiSine

RostyslavSavytskyy:
Added an option for modulating only with I or Q quadrature which gives us 2 sideband modulation. Also modified MultiSinePulseImplementation to the new format with the settings dictionary.

### DIFF
--- a/silq/instrument_interfaces/keysight/E8267D_interface.py
+++ b/silq/instrument_interfaces/keysight/E8267D_interface.py
@@ -185,6 +185,14 @@ class E8267DInterface(InstrumentInterface):
                                      'cannot be directly set, but is determined '
                                      'by FM_mode and whether pulses have '
                                      'frequency_sideband not None.')
+        self.add_parameter('IQ_channels',
+                           initial_value='IQ',
+                           vals=vals.Enum('IQ', 'I', 'Q'),
+                           set_cmd=None,
+                           docstring="Which channels to use for IQ modulation."
+                                     "Double-sideband modulation is used if only 'I' or 'Q' "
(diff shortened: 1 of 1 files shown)


silq/instrument_interfaces/keysight/E8267D_interface.py
```diff
@@ -671,7 +695,7 @@ def get_additional_pulses(self, interface: InstrumentInterface):
                             'input_instrument': interface.instrument_name(),
                             'input_channel': 'trig_in'})]
 
-        assert (interface.IQ_modulation() == 'on') and \
+        assert (IQ_modulation == 'on') and \
                (interface.FM_mode() == 'IQ'), 'FM_mode should be IQ and IQ_modulation should be ON ' \
```
maij, line 699: Is FM_mode part of `frequency_settings`? Or can you only access this via the interface?
I'm a bit out of the loop, I'm not sure when the interface was changed to pass more variables than just the interface...

maij (commented):
Looks good, I just have 1 question about how certain variables are passed to the pulse implementation functions.

silq/instrument_interfaces/keysight/E8267D_interface.py
RostyslavSavytskyy: No, so Serwan decided to detach some of the settings from the interface to make functions more independent, but not everything is in there, like FM_mode, etc. However, the interface still contains all of the parameters.

maij: Righto, I guess that's something that should be consolidated in future.

maij (approved):

PR #267, opened 2021-01-26
Fix use of matplotlib kwargs in fit toolbox

### LINKED ISSUE #206: Matplotlib kwarg aliases are used inconsistently and might cause errors in future
Basically, a newer version of maplotlib has deprecated the following usage and will instead raise a TypeError:
```python
x = np.linspace(0, 1, 10)
y = np.exp(x)
plt.plot(x, y, c='r', color='blue') # 'c' aliases to 'color' and is defined twice, this will raise an error
```

This might seem silly, but this can occur when we have plotting wrappers, e.g. in the [in the fit toolbox](https://github.com/nulinspiratie/SilQ/blob/master/silq/analysis/fit_toolbox.py#L27).

Where the default kwargs are:
`plot_kwargs = {'linestyle': '--', 'color': 'cyan', 'lw': 3}`

```python
 def add_to_plot(self,
                    ax: Axis,
                    N: int = 201,
                    xrange: Tuple[float] = None,
                    xscale: float = 1,
                    yscale: float = 1,
                    **kwargs):
        """Add fit to existing plot axis

            kwargs: Additional plot kwargs. By default Fit.plot_kwargs are used
```
---
and a bit later...
```python
        """
        plot_kwargs = {**self.plot_kwargs, **kwargs} # This can cause issues
        self.plot_handle, = ax.plot(
            x_vals_full, y_vals_full, **plot_kwargs
        )
        return self.plot_handle
```
---

Now I'm not sure how widespread this issue is, but there are tools available within matplotlib to clean up kwargs such as:
```python
>>> from matplotlib import cbook
>>> plot_kwargs = {'linestyle': '--', 'color': 'cyan', 'lw': 3}
>>> cbook.normalize_kwargs(plot_kwargs, alias_mapping=dict(linewidth=['lw']))
{'linewidth': 3, 'linestyle': '--', 'color': 'cyan'}
```

I would suggest at some point we look through how we handle plotting throughout silq and try to centralize it with `rcParams` or something.

maij:
Fixes #206

Still need to search through silq more broadly for places where we use matplotlib kwargs improperly.

### DIFF
--- a/silq/analysis/fit_toolbox.py
+++ b/silq/analysis/fit_toolbox.py
@@ -5,7 +5,6 @@ from lmfit.model import ModelResult
 
 from matplotlib import pyplot as plt
 from matplotlib.axis import Axis
-import matplotlib as mpl
 import matplotlib.cbook as cbook
 import matplotlib.lines as mlines
 
@@ -13,12 +12,6 @@ import logging
 from qcodes.data.data_array import DataArray
 
 logger = logging.getLogger(__name__)
-
-fit_rc_params = {
-    "axes.prop_cycle": mpl.cycler(color=['k']),
-    "lines.linestyle": "--",
-    "lines.linewidth": "2"
-}
 
 class Fit():
     """Base fitting class.
@@ -307,11 +300,12 @@ class Fit():
         x_vals_full *= xscale
         y_vals_full *= yscale
 
-        kwargs = cbook.normalize_kwargs(kwargs, mlines.Line2D)
-        plot_kwargs = self.default_plot_kwargs
-        plot_kwargs.update(**kwargs)
+        # Set default plot kwargs while de-aliasing (e.g. 'lw' -> 'linewidth')
+        # kwargs to prevent duplicate keys
+        kwargs = {**self.default_plot_kwargs,
+                  **cbook.normalize_kwargs(kwargs, mlines.Line2D)}
         self.plot_handle, = ax.plot(
-            x_vals_full, y_vals_full, **plot_kwargs
+            x_vals_full, y_vals_full, **kwargs
         )
         return self.plot_handle
 
--- a/silq/analysis/fit_toolbox.py
+++ b/silq/analysis/fit_toolbox.py
@@ -5,6 +5,7 @@ from lmfit.model import ModelResult
 
 from matplotlib import pyplot as plt
 from matplotlib.axis import Axis
+import matplotlib as mpl
 import matplotlib.cbook as cbook
 import matplotlib.lines as mlines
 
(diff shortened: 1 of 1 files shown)


silq/analysis/fit_toolbox.py
```diff
@@ -298,7 +306,10 @@ def add_to_plot(self,
             **{self.sweep_parameter: x_vals_full})
         x_vals_full *= xscale
         y_vals_full *= yscale
-        plot_kwargs = {**self.plot_kwargs, **kwargs}
+
+        kwargs = cbook.normalize_kwargs(kwargs, mlines.Line2D)
+        plot_kwargs = self.default_plot_kwargs
+        plot_kwargs.update(**kwargs)
```
nulinspiratie, line 312: Don't these two lines change `Fit.default_plot_kwargs`? I would use `plot_kwargs = {**self.default_plot_kwargs, **kwargs}`, which should prioritize kwargs from the second dict

nulinspiratie: and perhaps linewdith / lw should be treated separately... If a user also passes the kwarg `lw`, it would cause issues

nulinspiratie (changes requested):

silq/analysis/fit_toolbox.py
maij: Good catch, this is still a WIP, I was looking for a better solution than this. But I think I can combine things while maintaining clarity.

silq/analysis/fit_toolbox.py
```diff
@@ -298,9 +300,13 @@ def add_to_plot(self,
             **{self.sweep_parameter: x_vals_full})
         x_vals_full *= xscale
         y_vals_full *= yscale
-        plot_kwargs = {**self.plot_kwargs, **kwargs}
+
+        # Set default plot kwargs while de-aliasing (e.g. 'lw' -> 'linewidth')
+        # kwargs to prevent duplicate keys
+        kwargs = {**self.default_plot_kwargs,
+                  **cbook.normalize_kwargs(kwargs, mlines.Line2D)}
```
nulinspiratie, line 307: Wouldn't you want to apply normalize_kwargs on `{**self.default_plot_kwargs, **kwargs}`?

maij: No, that doesn't work. 

For example if you have:
```python
default_plot_kwargs = {'linewidth' : 2}
kwargs = {'lw': 3}

cbook.normalize_kwargs({**default_plot_kwargs, **kwargs}, mlines.Line2D)
```
This raises the following error:
```python-traceback
TypeError: Got both 'lw' and 'linewidth', which are aliases of one another
```

--- 

I could also normalize`self.default_plot_kwargs`, but I would like to think that people won't change the default arguments and pass in clashing kwargs too often...

nulinspiratie: Ah alright, didn't realize that normalize_kwargs couldn't handle that

nulinspiratie (approved):

PR #268, opened 2021-01-28
Force arbstudio to load unused channels with a DC pulse

### LINKED ISSUE #265: Instrument Interface: Arbstudio remains at whatever its last DC value when it was stopped
When the arbstudio is stopped it seems to remain playing whatever its last pulse was (I've observed it outputting a constant DC signal).

This is not generally a problem when not running measurements, but if you then target a pulse sequence that does not use this channel, the channel will continue outputting its last value while running a pulse sequence on the other channels.

I'm not sure what the best solution is for this, perhaps just always queing a zero-waveform for unused channels is the right way to go.

---

**Update**

This problem was even worse than previously thought, the arbstudio would output whatever pulse it was previously applying when triggering started again, even if the run command doesn't select those channels. That means you could have large sinusoids (in my case for adiabatic NMR, or in Rosty's case with microwave Q modulation) being applied when a different pulse sequence is loaded that doesn't use these channels.

maij:
Fixes #265.

More details pertaining to this issue:
@RostyslavSavytskyy and I did some testing and found that if you stopped the layout and started it again, the arbstudio would continue outputting on an unused channel, however if you stopped the layout, and explicitly started the triggering to the arbstudio (in this case the pulseblaster ESR Pro), then started the arbstudio, the unused channel would no longer output erroneously. 

Very strange! In any case, this seems to fix the issue.

### DIFF
--- a/silq/instrument_interfaces/lecroy/ArbStudio1104_interface.py
+++ b/silq/instrument_interfaces/lecroy/ArbStudio1104_interface.py
@@ -288,10 +288,13 @@ class ArbStudio1104Interface(InstrumentInterface):
         # that there are no times between pulses
         t_pulse = {ch: 0 for ch in self.active_channels()}
 
-        self.waveforms = {ch: [] for ch in self.active_channels()}
-        self.sequences = {ch: [] for ch in self.active_channels()}
+        # Unused channels load a single zero volt DC pulse sequence to ensure
+        # that no signal plays from these channels, this has minimal overhead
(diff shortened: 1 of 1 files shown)


nulinspiratie:
Hmm I wonder if it's something like if the ArbStudio stops before it reached the final pulse in it's waveform, it will somehow continue at that point.
However, if after it stops triggers are still being received such that it finished the current sequence, it will truly stop outputting anything

nulinspiratie (approved):
Yep I think this works

PR #272, opened 2021-02-08
Add setpoint_labels to DCSweepParameter

maij:
This is needed to prevent an error when using this parameter in a measurement context where setpoints are saved.


RostyslavSavytskyy:
Will test this soon ;)

RostyslavSavytskyy (approved):
Seems to be working

PR #274, opened 2021-02-22
fixes the issue when saving pulse traces as well

RostyslavSavytskyy:
In composite pulse sequences if you want to save also the traces of the applied pulses (plunge) and not just of the "read_initialize", the function (find_high_low) that determines the threshold voltage crushes due to the different sizes of the traces in each pulse. This modification quickly fixes the bug, but then the find_high_low function uses all traces to determine the threshold voltage which is useless. So we either need to determine the threshold voltage only when it is not set manually or be able to extract an use the traces only from read_initialize as it was done in non-composite parameters and pulse sequences.

**UPDATE**
It causes further crashes in the analysis, apparently composite analysis cannot handle the option when we save also traces that do not need to be analyzed.

### DIFF
--- a/silq/analysis/analysis.py
+++ b/silq/analysis/analysis.py
@@ -173,8 +173,8 @@ def find_high_low(
     if skip_pts > 0:
         traces = [trace[:, skip_pts:] for trace in traces]
 
-    # Turn list of 2D traces into a single 2D array
-    traces = np.ravel(traces)
+    # Turn list of 2D traces into a single 1D array
+    traces = np.concatenate([np.ravel(trace) for trace in traces])
 
     # Retrieve properties from config.analysis
     analysis_config = config.get("analysis", {})


maij:
Hi @RostyslavSavytskyy, sorry for the delay.

Thanks for raising this issue and attempting to solve it, however I don't really understand the changes you've made. By unravelling all of the traces you aren't actually removing the non-read pulses like you want to do. I've had a look through the code and have found a simple solution to this small bug, but I am wary that there may be a better long-term solution. See below for more details, my alternative fix will be contained in a code review.

---

I disagree that only looking at the read pulses is the best thing to do in all cases. There may be some merit in using an entire trace to determine threshold for a blip. Consider an empty-plunge-read sequence, where the SET current goes from high to low, then readout we should see blips (assuming we're doing compensated pulsing for now). Because this is measuring the actual SET current, it provides valuable information for determining the blip threshold automatically. The trouble we often see when looking at just the read section is that we are unable to histogram SET current correctly due to too few counts in the high current state. 

Overall, I think we may want to add an extra option (perhaps in `config.analysis.threshold_method`) to determine whether you want to use just the `read` sections of the pulse sequence, or the whole pulse sequence.

nulinspiratie:
Hi @RostyslavSavytskyy @maij 

I think the main consideration here is to keep the function `find_high_low` as simple as possible. I think the functionality it should have is: *receives (a list of) traces -> determines the high and low voltage of those combined traces*.
I therefore don't think it should care about what these traces actually are, e.g. if they are plunge or read pulses. This logic should be handled at a higher level.

My suggestion here is to filter out the read traces at the highest level possible.
So if it doesn't change the code too much, this should be done at the `AcquisitionParameterComposite.analyse` level, or else the `Analysis.analyse` level. If both those two would require significant changes it can be done even a level lower.

I do see Rosty's point that if the traces have different shapes, `np.ravel` won't work. Fixing this would indeed be an improvement, so I suggest a combination of Rosty's and Mark's suggestions:
```
if isinstance(traces, np.ndarray):
        traces = [traces]
# Turn list of 2D traces into a single 1D array
traces = np.concatenate([np.ravel(trace) for trace in traces])
```

RostyslavSavytskyy:
ok, I will apply the latest (Serwan's) version and test - will update on how it goes

RostyslavSavytskyy:
So far this seems to work on Tallulah, you can have a look and further test

maij:
@RostyslavSavytskyy Have you taken a look at @nulinspiratie's requested changes?

maij (approved):

PR #277, opened 2021-03-19
analyse_flips_old: Ensure results are wrapped in a list for consistency

maij:
This only applies to the old NMRParameter analysis, just uploading this so it doesn't go completely missing before we've finalised removing the non-composite parameters.


maij:
This code is no longer used. It ought to be removed in future.

PR #278, opened 2021-03-19
DCSweepParameter acquisition failing

### LINKED ISSUE #275: DCSweepParamater sometimes fails when performing 1D measurements
This is an issue that only seems to crop up under certain situations which I have yet to pin down.

When performing a measurement with a DCSweepParameter that sweeps 1 gate, the problem can arise that the expected shapes is not correct and can't be broadcast into the data array.

Once I have more information I will come back to this and fill this issue out a bit more clearly.
I believe it is related to the setpoints and may have been introduced from #272.

maij:
WIP

Attempting to fix #275


maij:
Addressed in #293.

PR #280, opened 2021-03-23
limited max amplitude and fixed triggering issues related to final delay

RostyslavSavytskyy:


silq/instrument_interfaces/keysight/Keysight_81180A_interface.py
```diff
@@ -512,7 +515,7 @@ def add_pulse(self, pulse, t, channel, sample_rate):
             )
 
         # Get waveform of current pulse
-        waveform = pulse.implementation.implement(sample_rate=sample_rate,)
+        waveform = pulse.implementation.implement(sample_rate=sample_rate, pulse=pulse)
```
nulinspiratie, line 518: This doesn't look right. Looking at all pulse implementations, it seems that DCPulseImplementation.implement is a staticmethod, while the rest isn't. So it looks like this line here would work for a DCPulse but not for other pulses. 

I think the correct solution is that DCPulseImplementation shouldn't have a staticmethod, though this might cause issues elsewhere

silq/instrument_interfaces/keysight/Keysight_81180A_interface.py
```diff
@@ -218,6 +219,8 @@ def setup(self, **kwargs):
         # targeted_pulse_sequence is the pulse sequence that is currently setup
         self.targeted_pulse_sequence = self.pulse_sequence
         self.targeted_input_pulse_sequence = self.input_pulse_sequence
+        assert self.pulse_sequence.final_delay >= 0.175e-6, \
+            f'Final pulse sequence delay {self.pulse_sequence.final_delay} must be >= 175ns.'
```
nulinspiratie, line 223: I can't remember if final_delay can also be zero or not. Does anyone else know?

nulinspiratie (dismissed):

silq/instrument_interfaces/keysight/Keysight_81180A_interface.py
RostyslavSavytskyy: Yes, the only reason I had to add this is that otherwise, it was giving me an error stating that it does not have the reference to which pulse it should use.

silq/instrument_interfaces/keysight/Keysight_81180A_interface.py
RostyslavSavytskyy: I tested this as well and I did not find any restrictions in the pulse_sequence.final_delay elsewhere. I added this restriction because during my tests if the final delay was smaller than this time, the oscilloscope could not properly trigger on the sequence, which seemed like there are some timing/trigger issues there, for arbstudio this time seems to be 211ns, which I haven't added yet, but probably should as well. At the moment for my setup, I set the default value to 320ns instead of 2ms, which is used on every other setup in the group. This 2ms time is by the way the reason we did not notice this pulseblaster delay issue earlier - the pre-delay in the first pulse was always exactly 2ms and was not defined by the parameters. I think my 320ns is defined based on the cable length/instruments delay, so doubt that it is general for everyone, but by setting the limitation to 175ns I wanted at least to warn everyone that there should be some limitations or at least it should be additionally tested.

nulinspiratie: Okay cool, yeah I think it makes sense that final delay must be larger than zero

silq/instrument_interfaces/keysight/Keysight_81180A_interface.py
nulinspiratie: I think the current change will not work for e.g. a SinePulse. I'll try to fix this when I have the time

RostyslavSavytskyy: Yes, can confirm this. It worked for DCPulse, but for SinePulse gave an error.

nulinspiratie: Try it now with the new commit I added

maij:
@RostyslavSavytskyy If this has been tested and is working, I'm happy to approve it as is.

maij (approved):

PR #282, opened 2021-05-06
Add FrequencyRampPulseImplementation to Keysight PXIe AWGs.

maij:
This implementation was missing so I've added it so we can do adiabatic sweeps.

### DIFF
--- a/silq/instrument_interfaces/keysight/Keysight_SD_AWG_interface.py
+++ b/silq/instrument_interfaces/keysight/Keysight_SD_AWG_interface.py
@@ -510,7 +510,7 @@ class SinePulseImplementation(PulseImplementation):
         sampling_rate = default_sampling_rate
         duration = self.pulse.duration
         # TODO: maybe make n_max an argument? Or even better: make max_samples a parameter?
-        waveform_multiple = 5  # the M3201A AWG needs the waveform length to be a multiple of 5
+        waveform_multiple = 5  # the Keysight SD AWG needs the waveform length to be a multiple of 5
         waveform_minimum = 15  # the minimum size of a waveform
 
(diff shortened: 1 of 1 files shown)


silq/instrument_interfaces/keysight/Keysight_SD_AWG_interface.py
```diff
@@ -647,6 +653,37 @@ def implement(self, interface, instrument, default_sampling_rate, threshold):
 
         return [waveform]
 
+class FrequencyRampPulseImplementation(PulseImplementation):
```
nulinspiratie, line 656: Would be good to have a check to see that the waveform isn't some crazy high amount of points, there's always the risk with ramp pulses

silq/instrument_interfaces/keysight/Keysight_SD_AWG_interface.py
```diff
@@ -647,6 +653,37 @@ def implement(self, interface, instrument, default_sampling_rate, threshold):
...
+        prescaler = 0 if sampling_rate == 500e6 else int(100e6 / sampling_rate)
+
+        samples = int(self.pulse.duration * sampling_rate)
+        samples -= samples % instrument.waveform_multiple
+        assert samples >= instrument.waveform_minimum, \
+            f"pulse {self.pulse} too short"
+
+        t_list = np.linspace(self.pulse.t_start, self.pulse.t_stop, samples)
+
+        waveform_data = self.pulse.get_voltage(t_list) / 1.5
```
nulinspiratie, line 672: Why /1.5? A comment would be helpful

nulinspiratie (approved):

PR #284, opened 2021-05-14
added buffer to have averaging over the window

RostyslavSavytskyy:
By waiting a specific number of scans, it will average over them and potentially improving the quality of the data

### DIFF
--- a/silq/tools/plot_tools.py
+++ b/silq/tools/plot_tools.py
@@ -837,9 +837,7 @@ class DCSweepPlot(ScanningPlot):
             initialize: Method called during initialization.
         """
         for k, name in enumerate(self.parameter.names):
-            if self.buffer_length > 1:
-                assert self.parameter.trace_pulse.enabled is False, \
-                    'Trace pulse must be turned off for running average.'
+            if self.buffer_length > 1 and name == 'DC_voltage':
                 self.buffers[self.buf_idx] = self.parameter.results[name]
                 self.buf_idx = (self.buf_idx + 1) % self.buffer_length
                 result = np.mean([buf for buf in self.buffers if buf is not None], axis=0)
--- a/silq/tools/plot_tools.py
+++ b/silq/tools/plot_tools.py
@@ -9,7 +9,6 @@ from time import time
 import numpy as np
 import logging
 
-
 import qcodes as qc
 from qcodes.plots.qcmatplotlib import MatPlot
 from qcodes.instrument.parameter import _BaseParameter
@@ -146,7 +145,7 @@ class MeasureSingle(PlotAction):
         self.txt = f"{self.plot.x_label}({event.xdata:.5f}\n" \
                    f"{self.plot.y_label}({event.ydata:.5g})\n" \
                    f"{self.plot.measure_parameter}_parameter.single_settings" \
-                   f"(samples={self.plot.samples_measure}, silent=False)\n"\
+                   f"(samples={self.plot.samples_measure}, silent=False)\n" \
                    f"qc.Measure({param}_parameter).run" \
                    f"(name='{self.plot.measure_parameter}_measure', quiet=True)"
         self.handle_code(self.txt, copy=True, execute=False)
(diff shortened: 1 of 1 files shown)


silq/tools/plot_tools.py
```diff
@@ -821,7 +833,15 @@ def update_plot(self, initialize=False):
             initialize: Method called during initialization.
         """
         for k, name in enumerate(self.parameter.names):
-            result = self.parameter.results[name]
+            if self.buffer_length > 1:
+                assert self.parameter.trace_pulse.enabled is False, \
+                    'Trace pulse must be turned off for running average.'
```
nulinspiratie, line 838: why is this?

silq/tools/plot_tools.py
```diff
@@ -791,10 +796,12 @@ class DCSweepPlot(ScanningPlot):
     """
     gate_mapping = {}
     point_color = 'r'
+
     # DCSweepParameter type
     def __init__(self,
                  parameter: _BaseParameter,
                  gate_mapping: Dict[str, str] = None,
+                 buffer_length: int = 1,
```
nulinspiratie, line 804: I don't think buffer_length is the best term for this. If the purpose is to average over multiple shots, I would call it something like "averages". The reason is that buffer_length can also be interpreted as "the length of a buffer", i.e. the number of trace points in a single buffer.

It would also be useful to have a docstring with the description of what this does

silq/tools/plot_tools.py
```diff
@@ -807,6 +814,11 @@ def __init__(self,
             subplots = 1
 
         self.point = None
+
+        self.buffer_length = buffer_length
+        self.buffers = [None, ] * self.buffer_length
+        self.buf_idx = 0
```
nulinspiratie, line 820: Shouldn't buf_idx be reset in update_plot when initialize=False?

silq/tools/plot_tools.py
```diff
@@ -821,7 +833,15 @@ def update_plot(self, initialize=False):
             initialize: Method called during initialization.
         """
         for k, name in enumerate(self.parameter.names):
-            result = self.parameter.results[name]
+            if self.buffer_length > 1:
+                assert self.parameter.trace_pulse.enabled is False, \
+                    'Trace pulse must be turned off for running average.'
+                self.buffers[self.buf_idx] = self.parameter.results[name]
+                self.buf_idx = (self.buf_idx + 1) % self.buffer_length
+                result = np.mean([buf for buf in self.buffers if buf is not None], axis=0)
```
nulinspiratie, line 841: Does this still return a measurement every acquisition, or only once every `buffer_length` acquisitions? It looks like it's returning every acquisition

nulinspiratie (changes requested):

silq/tools/plot_tools.py
RostyslavSavytskyy: I see, ok, can rename

silq/tools/plot_tools.py
RostyslavSavytskyy: it is reset there as well, but I see your point of no need to define it here as well then, can correct.

nulinspiratie: Ah no if it's reset somewhere else then keep it here.
Generally it's good practice to define all class attributes in the __init__

silq/tools/plot_tools.py
RostyslavSavytskyy: Because if `self.parameter.trace_pulse.enabled is True`, then `self.parameter.results[name]` contains also data from trace_pulse and it complicates implementation of this "averaging thing". I agree this is not a neat solution, can try to improve

silq/tools/plot_tools.py
RostyslavSavytskyy: yes, every acquisition - the idea is to see yourself how the data quality improves with time the longer you wait

silq/tools/plot_tools.py
RostyslavSavytskyy: ok

maij:
@RostyslavSavytskyy Cool, I'll see if I can spare some time for this one.

maij (approved):

PR #285, opened 2021-05-14
space correction for gen tools function

RostyslavSavytskyy:


nulinspiratie:
I don't think this is an improvement. It should iterate over all possible points, and only return None if all of them are unsuccessful.
Have a look at the ["for ... else ..." structure](https://book.pythontips.com/en/latest/for_-_else.html)

RostyslavSavytskyy:
I see, did't know that, thought it was a spacing mistake... cool.

Issue #290, opened 2021-10-07
Plotting improvements


maij:
I am just making a list of basic improvements for our plotting routines that would help tremendously.

# Pulse sequence plotting
When we plot pulse sequences we could add a 'schematic' alternative similar to what Serwan cooked up in his thesis.
This would make intra-group communications far more effective when people can see what pulses we are applying.
![Screenshot from 2021-10-07 15-56-35](https://user-images.githubusercontent.com/7358024/136322967-2f6b83b9-daeb-45c2-b300-dd6d12583934.png)

#  MatPlot handling of 2D data that isn't meshed.
Sometimes we want to plot 2D data where the x-axis/y-axis isn't continguously spaced. This causes the squares to all be misshapen, I think we can easily come up with an alternative.

PR #291, opened 2021-10-08
Addition of a schematic plotting method for pulse sequences

maij:
This address issue  a plotting problem I raised in #290.

This PR adds a `plot_schematic` method to pulse sequences to easily show a pictorial representation of a pulse sequence. It's still in Beta, I would encourage everyone to try it out on their sequences and suggest improvement (and contribute improvements if you can!)


HollySte:
Doesn't work for pulses that have a power and no amplitude.

HollySte:
Also the read level is hard-coded at 0V innit

PR #292, opened 2021-10-11
Enable single-writer multiple-reader access to trace files.

### LINKED ISSUE #224: Reading an hdf5 file while it is being written can corrupt the data
There have been instances in the past where a process performing a measurement will be writing to an hdf5 file, while a user attempts to load the file from another process to perform analysis. In this case, there is a possibility that the reader will somehow cause the data to become corrupted.

A potential solution to this is to move all hdf5 file use towards the [single-writer multiple-reader](http://docs.h5py.org/en/stable/swmr.html) framework, which addresses this issue directly. The main drawback of this is that older version of hdf5 (h5py) will be unable to load these new files, but that shouldn't be a problem as the current required version of h5py is >= 2.6 (which includes SWMR).

This was originally flagged in #217.

maij:
Fixes #224.

By using the latest version of hdf5 we gain access to some extra features, including single-writer multiple-reader (SWMR) mode.
We may wish to specify the exact version we want to use, going forward, to ensure backwards compatibility is more easily assured.

- [x] Test that I can read from a live trace file as it changes.
- [x] Verify the integrity of the traces (compare locally acquired version to version recovered from file)
- [x] I also checked that loading h5py files from previous library versions works well.
- [x] Speed performance, actually seems to have improved
    - Before
        - Storing 600,000 points takes ~ 260 +- 58 ms.
        - Storing 1500 points takes ~5.7 +- 4.7 ms
    - After 
        - Storing 600,000 points takes ~ 230 +- 50 ms.
        - Storing 1500 points takes ~1 +- 4 ms

### DIFF
--- a/silq/meta_instruments/layout.py
+++ b/silq/meta_instruments/layout.py
@@ -1658,7 +1658,7 @@ class Layout(Instrument):
         # Create new hdf5 file
         filepath = os.path.join(folder, f'{name}.hdf5')
         assert not os.path.exists(filepath), f"Trace file already exists: {filepath}"
-        file = h5py.File(filepath, 'w')
+        file = h5py.File(filepath, 'w', libver='latest')
 
         # Save metadata to traces file
         file.attrs['sample_rate'] = self.sample_rate()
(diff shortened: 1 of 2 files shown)


nulinspiratie (approved):
Great addition! Do we need a minimal h5py version?

maij:
> Great addition! Do we need a minimal h5py version?

Nope! The current minimum in `setup.py` is 2.6.0 which supports SWMR.

PR #295, opened 2022-01-12
Add auto-tuning routines

maij:
WIP. Input is welcome!

This PR integrates the autotuning scripts developed by Serwan for various experiments. It employs the Nelder-mead optimisation algorithm provided by scipy.

Todo:
- [ ] Improve code clarity
- [ ] Improve ease-of-use, users should be easily able to provide some parameters to optimize some function over (typically gate voltages, but could be something else such as pulse duration) and a simple cost function that is evaluated.
- [ ] Create an abstract class for cost evaluation function. (Not sure about this, it is currently written such that you just need to provide a function that gives you the cost at some tuning point)

At the moment this module has no classes, and it might not need them. We don't have any objects that need to be passed around so this might be fine.

### DIFF
--- a/silq/tools/autotuning.py
+++ b/silq/tools/autotuning.py
@@ -1,0 +1,242 @@
+"""Functions for auto-tuning using the Nelder Mead optimization algorithm"""
+
+from scipy import optimize
+from qcodes import Sweep
+
+
+def _create_initial_simplex(initial_values: dict, delta: float = None, **kwargs):
+    if delta is None:
+        assert len(kwargs) == len(initial_values)
+
+    x0 = list(initial_values.values())
+    initial_simplex = [x0]
+    for k, name in enumerate(initial_values):
+        x0_copy = x0[:]
+        x0_copy[k] += kwargs.get(name, delta)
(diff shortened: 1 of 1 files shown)


nulinspiratie:
Hey @maij  good suggestion!
I think your second todo item is already covered by `optimize_nelder_mead`, don't you think? My thought was to make a general function that only needs a cost function and then optimizes over it. So it isn't specific to gate voltages.
I also think that an abstract class isn't needed. The requirements for a cost function are super general: `provide values for variables` -> `return cost`.

I think it makes sense to split the file into two: one contains the general optimization functions (currently the top two functions) and another that is specific for voltage optimization etc. (the rest of the file)

Also, when I was writing this code, I was uncertain whether I should use classes or just functions. In the end I went for functions because I thought the `optimize_nelder_mead` would be very general, whereas the functions calling this would vary a lot in how they're structured. But it's something I'm not entirely sure about, what are your thoughts?

Once we have converged here I can at least give a crack at writing docstrings for the methods, that will hopefully make their usage more clear

